PR #1, opened 2020-07-08
auto generate name

yarongol:


moshe-kabala:
@omer-dayan  Already implement it

PR #2, opened 2020-07-08
wait on job

yarongol:


moshe-kabala:
Already done as a part of RUN-965

PR #14, opened 2020-08-08
cluster awareness

yarongol:


cmd/arena/commands/cluster/set.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+
+	"github.com/kubeflow/arena/pkg/util/command"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func runSetCommand(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
```
ocherfas, line 12: This can be done using an argument on 
```
cobra.Cobra{
Args:  cobra.RangeArgs(1, 1),
}
```

PR #22, opened 2020-09-24
RUN-967 - adding table interface to utils, some functions to run a query in prom and adding (nornmolize) the top node

moshe-kabala:


pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	for queryName, query := range q {
+		getFunc := func() error {
+			rst, err := ps.Query(query)
+			mux.Lock()
+			queryResults[queryName] = rst
+			// fmt.Print(queryName,"\n\n",rst, "\n\n")
+			mux.Unlock()
+			return err
+		}
+		getFunc();
```
yodarshafrir1, line 152: did you mean to use goroutine?

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	return
+ }
+
+
+ // MultipuleQueriesToItemsMap map multipule queries to items by given itemId
+func (ps *Client) MultipuleQueriesToItemsMap(q MultiQueries, itemID string) ( ItemsMap, error) {
+	queryResults := map[string]MetricData{}
+	rst := ItemsMap{}
+	// funcs := []func() error{}
+	var mux sync.Mutex
```
yodarshafrir1, line 139: why do you need mutex here? they all update a different value

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	// err = util.Parallel(funcs...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// map the result to items by the given 'itemId' 
+	for queryName, qr := range queryResults {
+		// todo: check the metric type
```
yodarshafrir1, line 163: do we want this todo here?

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+		for _, metricResult := range qr.Result {
+			// search the itemId in metric labels
+			key, ok := metricResult.Metric[itemID]
+			if !ok {
+				return nil, fmt.Errorf("[Prometheos] Not found an 'itemID' (%s) on the metric query: %s => %s",itemID, queryName, q[queryName] )
```
yodarshafrir1, line 168: This error will later be shown to the user.
the user has no idea what is `itemID`. let's rephrase it

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+			val := metricResult.Value
+			item, created := rst[key]
+			if !created {
```
yodarshafrir1, line 172: can it already be created? if so please explain

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+
+func (td *tableData) addFields(modelType reflect.Type, path []string, groupTag GroupTag, showByDefult bool) {
+	fieldsCount := modelType.NumField()
+	for i := 0; i < fieldsCount; i++ {
+		td.addField( modelType.Field(i), path, groupTag, showByDefult)
+	}
+}
+
+func (td *tableData) addField(fieldType reflect.StructField, path []string, groupTag GroupTag, showByDefult bool) {
+	// if need to hide the filed
```
yodarshafrir1, line 147: filed -> field

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+
+	}
+	return td
+}
+
+func (td *tableData) Error() error {
+	return td.err
+}
+
+//// helpers
```
yodarshafrir1, line 303: let's extract to a new file called helpers?

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+	}, nil
+}
+
+func interfaceToArrayOfInterface(a interface{}) ([]interface{}, error) {
+	object := reflect.ValueOf(a)
+	items := make([]interface{}, object.Len())
+	for i := 0; i < object.Len(); i++ {
+		items[i] = object.Index(i).Interface()
+	}
+	return items, nil
```
yodarshafrir1, line 362: if it never returns an error why do we want to have this in the type of the return value

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// UnwrapValuePtr recursively unwrap pointers
+func UnwrapValuePtr(ft reflect.Value) *reflect.Value {
```
yodarshafrir1, line 366: why ft?

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+		if ft.IsNil() {
+			return nil
+		}
+		return UnwrapValuePtr(ft.Elem())
+	}
+	return &ft
+}
+
+// UnwrapTypePtr recursively unwrap pointers
+func UnwrapTypePtr(ft reflect.Type) reflect.Type {
```
yodarshafrir1, line 377: why ft

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+		return ft.String()
+	case reflect.Float32, reflect.Float64:
+		return fmt.Sprintf("%.2f", ft.Float())
+	case reflect.Ptr:
+		if ft.IsNil() {
+			return ""
+		}
+		return StringifyValue(ft.Elem())
+	default:
+		// todo
```
yodarshafrir1, line 399: todo what?

pkg/ui/table.go
```diff
@@ -0,0 +1,434 @@
...
+		return string(ft.String())
+	}
+}
+
+
+func getNesstedVal(t reflect.Value, path []string) (val *reflect.Value) {
```
yodarshafrir1, line 405: why t

pkg/ui/table_tes.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type Person struct {
+	name string `title:"Name" def:"--"`
+	Age int
+	address Address `group:"Addres"`
+	addressPtr *Address `group:"Addres2"`
+}
+
+func RunTableExampleTes() {
```
yodarshafrir1, line 21: Please rename this and task, you are missing a `t` in `test`

yodarshafrir1: please use the golang test framework so it will run in ci as well?

pkg/util/cache.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	return &cache
+}
+
+func (c *cacheImpl) wait() (interface{}, error) {
+	return (<- c.waitChan)()
+}
+
+func (c *cacheImpl) Get() (interface{}, error) {
+	c.mux.Lock()
```
yodarshafrir1, line 48: why is the mutex needed?

pkg/util/cache.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (c *cacheImpl) Status() CacheStatus {
+	return c.status
+}
+
+
+func (c *cacheImpl) Refresh() (interface{}, error) {
+	c.mux.Lock()
```
yodarshafrir1, line 73: why is the mutex needed

pkg/util/cache.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type cacheImpl struct {
+	get_data GetCacheDataFunc
+	data interface{}
+	err error
+	status CacheStatus
+	waitChan chan GetCacheDataFunc
+	waitersCounter int
+	mux sync.Mutex
```
yodarshafrir1, line 31: please rename to mutex

pkg/util/parallel.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+
+// Parallel - running multipule functions in parallel
```
yodarshafrir1, line 4: multiple

yodarshafrir1: this method isnt used - please delete this file

yodarshafrir1 (changes requested):
Reviewed, please see my notes.
Also please do the following:
1. Remove all the functions from the types.go file
2. Remove all irrelevant `todo`s along the code.
3. Check the output of this command with fractional GPUs jobs, and then without GPU nodes at all (you can just disable the nvidia-device-plugin, I can help you with that).
4. Please rebase

cmd/services/node.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func NewNodeDescriber(client kubernetes.Interface, pods []v1.Pod) *NodeDescriber {
+	return &NodeDescriber{
+		client:  client,
+		allPods: pods,
+	}
+}
+
+func (d *NodeDescriber) GetAllNodeInfos() ([]t.NodeInfo, error, error) {
```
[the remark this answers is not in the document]
moshe-kabala: Fixed

cmd/services/node.go
```diff
@@ -0,0 +1,80 @@
...
+
+	var promData prom.ItemsMap
+
+	// get prometheus node resources data
+	promClient, err := prom.BuildPromethuseClient(d.client)
+	if err == nil {
+	
+		promData, err = promClient.MultipuleQueriesToItemsMap(nodePQs, promethesNodeLabelID)
+		
+	}
```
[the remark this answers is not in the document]
moshe-kabala: I didn't use else to cover the case of the internal if code return err too

cmd/services/node.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		warn = fmt.Errorf("Missing some data. \nresone: Can't access to the prometheus server, \ncause err: %s", err)
```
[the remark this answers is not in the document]
moshe-kabala: fix the typo but the 'n' belong to \n (newline).

cmd/services/node.go
```diff
@@ -0,0 +1,80 @@
+package services
+
+import (
+	"fmt"
+
+	t "github.com/run-ai/runai-cli/cmd/types"
```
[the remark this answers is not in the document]
moshe-kabala: fixed

cmd/services/node.go
```diff
@@ -0,0 +1,80 @@
...
+			promData,
+		)
+		nodeInfoList = append(nodeInfoList, nodeInfo)
+	}
+	return nodeInfoList, nil, warn
+}
+
+func (d *NodeDescriber) GetPodsFromNode(node v1.Node) []v1.Pod {
+	pods := []v1.Pod{}
+	for _, pod := range d.allPods {
```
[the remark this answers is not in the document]
moshe-kabala: Good Question. this is an old code I don't touch the logic here
may be the right place to filter these pods is here.
what do you think?

cmd/services/node.go
```diff
@@ -0,0 +1,80 @@
...
+		promData, err = promClient.MultipuleQueriesToItemsMap(nodePQs, promethesNodeLabelID)
+		
+	}
+	if err != nil {
+		warn = fmt.Errorf("Missing some data. \nresone: Can't access to the prometheus server, \ncause err: %s", err)
+	}
+
+	for _, node := range nodeList.Items {
+		pods := d.GetPodsFromNode(node)
+		nodeInfo := t.NewNodeInfo(
```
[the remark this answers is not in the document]
moshe-kabala: I have been debating a lot with myself what to put in there maybe you have some rule of thumb
I put the constructor there because of private fields (I change them to public latter)
I fell that the type and its constructor should be in the same package. and maybe this type need to be in services ?

cmd/tensorboard.go
```diff
@@ -135,3 +135,26 @@ func tensorboardURL(name, namespace string, clientset kubernetes.Interface) (url
 
 	return url, nil
 }
+
+// todo: create kube utils
+func isNodeReady(node v1.Node) bool {
```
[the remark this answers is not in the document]
moshe-kabala: Fixed

cmd/tensorboard.go
```diff
@@ -135,3 +135,26 @@ func tensorboardURL(name, namespace string, clientset kubernetes.Interface) (url
...
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+// todo: create kube utils
+func getNodeInternalAddress(node v1.Node) string {
```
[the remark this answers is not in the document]
moshe-kabala: Fixed

cmd/tensorboard.go
```diff
@@ -135,3 +135,26 @@ func tensorboardURL(name, namespace string, clientset kubernetes.Interface) (url
...
+	address := "unknown"
+	if len(node.Status.Addresses) > 0 {
+		//address = nodeInfo.node.Status.Addresses[0].Address
+		for _, addr := range node.Status.Addresses {
+			if addr.Type == v1.NodeInternalIP {
+				address = addr.Address
```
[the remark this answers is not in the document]
moshe-kabala: yes, changed (this is an old code I don't notice to the implement)

cmd/tensorboard.go
```diff
@@ -135,3 +135,26 @@ func tensorboardURL(name, namespace string, clientset kubernetes.Interface) (url
...
+			}
+		}
+	}
+	return address
```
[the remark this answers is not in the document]
moshe-kabala: changed

cmd/types/node.go
```diff
@@ -0,0 +1,130 @@
...
+
+type ClusterNodesView struct {
+	GPUs            	int
+	UnhealthyGPUs   	int
+	AllocatedGPUs   	float64
+	GPUsOnReadyNode 	int
+}
+
+
+func (cnv *ClusterNodesView) Render(w io.Writer) {
```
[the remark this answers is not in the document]
moshe-kabala: Why ?
where you think I need to put it?

cmd/types/pod.go
```diff
@@ -0,0 +1,39 @@
...
+
+type PodResourcesStatus struct {
+	Limited ResourceList
+	Allocated ResourceList
+	Requested ResourceList
+	Usage ResourceList
+}
+
+
+func GetPodResourceStatus(pod v1.Pod) PodResourcesStatus {
```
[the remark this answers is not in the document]
moshe-kabala: yes, it can belong to a converters package. but I don't have so many converters to create a separate package 
I wanted to put it in util but it created circle dependency so I put it there for now. 
do you think a new converters package is a good idea?

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if err != nil {
+		return 
+	} else if len(list.Items) > 0 {
+		service = &list.Items[0]
+	} else {
+		return nil, fmt.Errorf("Not found a server for prometheus")
+	}
+
+	return
```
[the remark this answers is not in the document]
moshe-kabala: changed (remained the if > else instead if  > if)

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	list, err := ps.client.CoreV1().Services(namespace).List( metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("app=%s", promLabel),
+	})
+
+	if err != nil {
+		return 
+	} else if len(list.Items) > 0 {
+		service = &list.Items[0]
+	} else {
+		return nil, fmt.Errorf("Not found a server for prometheus")
```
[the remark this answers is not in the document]
moshe-kabala: fixed

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		log.Debugf("Query prometheus failed due to err %v", err)
+		log.Debugf("Query prometheus failed due to result %s", string(metric))
+		return
+	}
+	rst = &Metric{}
+	err = json.Unmarshal(metric, rst)
+	log.Debugf("Prometheus metric:%v", rst)
+	if err != nil {
+		err = fmt.Errorf("failed to unmarshall heapster response: %v", err)
```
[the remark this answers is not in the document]
moshe-kabala: fixed

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if rst.Status != SuccessStatus {
+		err = fmt.Errorf("failed to query prometheus, status: %s", rst.Status)
+		return 
+	}
+	if len(rst.Data.Result) == 0 {
+		log.Debugf("The metric is not exist in prometheus for query  %s", query)
+	}
+	data = rst.Data
+	return
```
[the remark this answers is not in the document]
moshe-kabala: this return the predefined return values in the signature of the function

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+ }
+
+
+ // MultipuleQueriesToItemsMap map multipule queries to items by given itemId
+func (ps *Client) MultipuleQueriesToItemsMap(q MultiQueries, itemID string) ( ItemsMap, error) {
+	queryResults := map[string]MetricData{}
+	rst := ItemsMap{}
+	// funcs := []func() error{}
```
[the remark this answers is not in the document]
moshe-kabala: I make it in comment because of a bug in the parallel requests to Prometheus. I wish to fix this bug somewhere  ( maybe when we will decide to improve the performance of any command that will use multiple queries vs Prometheus) and then return the comment code - for now, I can leave there a comment. it is OK?

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	var mux sync.Mutex
+
+	var err error;
+
+	for queryName, query := range q {
+		getFunc := func() error {
+			rst, err := ps.Query(query)
+			mux.Lock()
+			queryResults[queryName] = rst
+			// fmt.Print(queryName,"\n\n",rst, "\n\n")
```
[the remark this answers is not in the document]
moshe-kabala: done

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+			mux.Unlock()
+			return err
+		}
+		getFunc();
+		// funcs = append(funcs, getFunc)
```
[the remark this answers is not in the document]
moshe-kabala: done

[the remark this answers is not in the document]
moshe-kabala: the same as above commit

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	data = rst.Data
+	return
+ }
+
+
+ // MultipuleQueriesToItemsMap map multipule queries to items by given itemId
+func (ps *Client) MultipuleQueriesToItemsMap(q MultiQueries, itemID string) ( ItemsMap, error) {
+	queryResults := map[string]MetricData{}
+	rst := ItemsMap{}
```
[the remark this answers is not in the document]
moshe-kabala: Do you have any idea?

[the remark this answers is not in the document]
moshe-kabala: how about GroupMultipuleQueriesToItemsMap ?

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+			mux.Lock()
+			queryResults[queryName] = rst
+			// fmt.Print(queryName,"\n\n",rst, "\n\n")
+			mux.Unlock()
+			return err
+		}
+		getFunc();
+		// funcs = append(funcs, getFunc)
+	}
+	// err = util.Parallel(funcs...)
```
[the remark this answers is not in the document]
moshe-kabala: the same as above commit

pkg/prometheus/prometheus.go
moshe-kabala: I want to use goroutine by the parallel function but as I wrote I have a weird bug there (the response overwrite by anther response) maybe you can help me with that

pkg/prometheus/prometheus.go
moshe-kabala: removed

pkg/ui/table.go
moshe-kabala: because copy past from getNesstedType :) fixed

pkg/ui/table.go
moshe-kabala: just I know :). removed

pkg/ui/table.go
moshe-kabala: field type 
change to t (type)

pkg/ui/table.go
moshe-kabala: field type
change to v (value)

pkg/ui/table.go
moshe-kabala: I forgat to cover the case that the value is not iterable, added

pkg/ui/table.go
moshe-kabala: some of the helpers belong only to the table some are not. 
I created a file for the general helpers
for table-specific helpers I can create another file table_helpers or remain them in the table. what do you think?

pkg/ui/table.go
moshe-kabala: fixed

pkg/prometheus/prometheus.go
moshe-kabala: Yes, there are multi-queries and each one contains multi-metrics. the item created dynamically by iterating  over the metrics so the first time that the item id appear in a metric we need to create the item instance

pkg/prometheus/prometheus.go
moshe-kabala: first, the error not will happen even though the Prometheus is not available (just if we have a problem with the metrics)
the error will be wrapped in a nice message that Prometheus service not found 
I assume that if this error happens we want the message to understand what went wrong
do you agree?

pkg/prometheus/prometheus.go
moshe-kabala: yes for now we query only vector but in the latter, we will query also matrix
I can return an error if the type is not a vector for now

pkg/ui/table_tes.go
moshe-kabala: Done

pkg/util/cache.go
moshe-kabala: In the beginning, the cache design to be able to save async data in lazy mode (request the data only if someone ask it) and provide it to any goroutine thread 
for example, we have the following function 
```
func GetSomeData(id string) Data {
   client := clientCache.Get()
   return client.getDataById(id)
}
```

and in another area, we call it in parallel
```
go GetSomeData(id1)
go GetSomeData(id2)
go GetSomeData(id3)
```

the cache will keep that the function "clientCache.Get()" will be call to the data only once
so this is the reason for mutex in the cache when I change any of the cache data
but for now, I don't use it so I can remove it and keep it aside or remained it there

pkg/util/cache.go
moshe-kabala: see comment above

pkg/util/cache.go
moshe-kabala: done

cmd/services/node.go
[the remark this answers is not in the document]
moshe-kabala: For now, I filter out all the jobs that are not in running mode, and if the node is not ready I return empty array

cmd/services/node.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func NewNodeDescriber(client kubernetes.Interface, pods []v1.Pod) *NodeDescriber {
+	return &NodeDescriber{
+		client:  client,
+		allPods: pods,
+	}
+}
+
+func (d *NodeDescriber) GetAllNodeInfos() ([]types.NodeInfo, error, string) {
```
yodarshafrir1, line 37: The convention in go is to put the `error` last

cmd/services/node.go
```diff
@@ -0,0 +1,82 @@
...
+	nodeList, err := d.client.CoreV1().Nodes().List(metav1.ListOptions{})
+
+	if err != nil {
+		return nodeInfoList, err, warn
+	}
+
+	var promData prom.ItemsMap
+
+	// get prometheus node resources data
+	promClient, err := prom.BuildPromethuseClient(d.client)
```
yodarshafrir1, line 50: promethuse - > prometheus

cmd/services/node.go
```diff
@@ -0,0 +1,82 @@
...
+
+func NewNodeDescriber(client kubernetes.Interface, pods []v1.Pod) *NodeDescriber {
+	return &NodeDescriber{
+		client:  client,
+		allPods: pods,
+	}
+}
+
+func (d *NodeDescriber) GetAllNodeInfos() ([]types.NodeInfo, error, string) {
+	var warn string
```
yodarshafrir1, line 38: rename to `warning`

cmd/services/node.go
```diff
@@ -0,0 +1,82 @@
...
+	return nodeInfoList, nil, warn
+}
+
+func (d *NodeDescriber) GetPodsFromNode(node v1.Node) []v1.Pod {
+	pods := []v1.Pod{}
+	if !util.IsNodeReady(node) {
+		return pods
+	}
+	for _, pod := range d.allPods {
+		if pod.Spec.NodeName == node.Name && pod.Status.Phase == v1.PodRunning{
```
yodarshafrir1, line 76: Running and Pending.
Pending pods can be allocated to nodes as well, they can every get stuck on Pending state and still request the Node's resources

cmd/util/node.go
```diff
@@ -0,0 +1,99 @@
...
+		case k == nodeLabelRole && v != "":
+			roles.Insert(v)
+		}
+	}
+	return roles.List()
+}
+
+func GetNodeInternalAddress(node v1.Node) string {
+	if len(node.Status.Addresses) > 0 {
+		//address = nodeInfo.node.Status.Addresses[0].Address
```
yodarshafrir1, line 82: Please delete the comment

cmd/util/node.go
```diff
@@ -0,0 +1,99 @@
...
+		for _, addr := range node.Status.Addresses {
+			if addr.Type == v1.NodeInternalIP {
+				return addr.Address
+			}
+		}
+	}
+	return "unknown"
```
yodarshafrir1, line 89: https://github.com/run-ai/runai-cli/pull/22#discussion_r500796301

cmd/types/node.go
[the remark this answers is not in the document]
yodarshafrir1: The package name is `types`, method with the name `Render` doesn't belong here.
You can create a new file under utils

cmd/types/node.go
```diff
@@ -0,0 +1,130 @@
...
+			gpuUnhealthyPercentage = float64(cnv.UnhealthyGPUs) / float64(cnv.GPUs) * 100
+		}
+		fmt.Fprintf(w, "%s/%s (%d%%)\t\n",
+			strconv.FormatInt(int64(cnv.UnhealthyGPUs), 10),
+			strconv.FormatInt(int64(cnv.GPUs), 10),
+			int64(gpuUnhealthyPercentage))
+	}
+}
+
+func (cnv *ClusterNodesView) AddNode(status NodeStatus, gpu NodeGPUResource) {
```
yodarshafrir1, line 122: Same comment as above

cmd/types/helpers.go
```diff
@@ -0,0 +1,61 @@
+package types
+
+import (
+	"strconv"
+
+	v1 "k8s.io/api/core/v1"
+	prom "github.com/run-ai/runai-cli/pkg/prometheus"
+)
+
+func setIntPromData(num *int64, m map[string][]prom.MetricValue, key string) error {
```
yodarshafrir1, line 10: all these method shouldn't be under types

cmd/types/pod.go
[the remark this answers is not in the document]
yodarshafrir1: It's ok to create a new package even if it will only 1 converter.
Looks like it belongs utils the most.
the reason you have circle dependency is because you put `func (ni *NodeInfo) GetResourcesStatus() NodeResourcesStatus` under types as well, move them both from this package and it will be fine

pkg/prometheus/prometheus.go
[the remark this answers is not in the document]
yodarshafrir1: could work, also `queriesToResponse` would also work.
please also rename the `rst`

pkg/prometheus/prometheus.go
yodarshafrir1: wasn't removed.
You don't need mutex here

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+		funcs = append(funcs, getFunc)
+	}
+	err := util.Parallel(funcs...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// map the result to items by the given 'itemId' 
+	for queryName, qr := range queryResults {
```
yodarshafrir1, line 162: You should wait for the result

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+	rst := ItemsMap{}
+	funcs := []func() error{}
+	var mux sync.Mutex
+	for queryName, query := range q {
+		// create scoped vars for the function
+		query := query
+		name := queryName
+		getFunc := func() error {
+			rst, err := ps.Query(query)
+			mux.Lock()
```
yodarshafrir1, line 146: No need for mutex.
every query writes to a different key

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+			queryResults[name] = rst
+			mux.Unlock()
+
+			return err
+		}
+	
+		funcs = append(funcs, getFunc)
+	}
+	err := util.Parallel(funcs...)
```
yodarshafrir1, line 155: Please don't use this generic method for goroutines, just make the call here

pkg/prometheus/prometheus.go
yodarshafrir1: So write a more detailed todo please

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	// map the result to items by the given 'itemId' 
+	for queryName, qr := range queryResults {
+		// todo: check the metric type
+		for _, metricResult := range qr.Result {
+			// search the itemId in metric labels
```
yodarshafrir1, line 165: please remove the comment, it's obvious

pkg/prometheus/prometheus.go
yodarshafrir1: The problem is with the word `itemID`.
You can write 
```suggestion
				return nil, fmt.Errorf("[Prometheos] Failed to find key: (%s) on the metric query: %s => %s",itemID, queryName, q[queryName] )
```

pkg/util/cache.go
yodarshafrir1: If you don't use the mutex please delete it

cmd/services/node.go
moshe-kabala: done

cmd/services/node.go
moshe-kabala: done

cmd/services/node.go
moshe-kabala: done

cmd/services/node.go
moshe-kabala: good to know, 
done

yodarshafrir1 (changes requested):
Let's talk about this refactor

cmd/util/node.go
moshe-kabala: done

pkg/prometheus/prometheus.go
moshe-kabala: The advantage of this util method is it makes the logic of waiting for you

pkg/prometheus/prometheus.go
moshe-kabala: done

cmd/get/get.go
```diff
@@ -108,9 +107,9 @@ type PrintArgs struct {
 /*
 * search the training job with name and training type
  */
-func searchTrainingJob(kubeClient *client.Client, jobName string, trainingType string, namespaceInfo cmdTypes.NamespaceInfo) (job TrainingJob, err error) {
+func SearchTrainingJob(kubeClient *client.Client, jobName string, trainingType string, namespaceInfo cmdTypes.NamespaceInfo) (job trainer.TrainingJob, err error) {
```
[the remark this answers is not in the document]
moshe-kabala: It used by exec and logs packages. it better to put it in the util but I don't do that refactor yet

cmd/job/delete/delete.go
```diff
@@ -79,14 +82,14 @@ func deleteTrainingJob(kubeClient *client.Client, jobName string, namespaceInfo
 
 	// 2. Handle training jobs created by arena
 	if trainingType == "" {
-		trainingTypes, err = getTrainingTypes(jobName, namespaceInfo.Namespace, kubeClient.GetClientset())
+		trainingTypes, err = get.GetTrainingTypes(jobName, namespaceInfo.Namespace, kubeClient.GetClientset())
```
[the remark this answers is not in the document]
moshe-kabala: yes agree. but I must to that to separate them first the seconde step is to move that logic into share util package

pkg/prometheus/prometheus.go
moshe-kabala: the parallel function do it for me

pkg/prometheus/prometheus.go
moshe-kabala: removed

cmd/types/node.go
moshe-kabala: All the function in types passed to a helpers package

cmd/types/helpers.go
moshe-kabala: All the function in types passed to a helpers package

pkg/prometheus/prometheus.go
moshe-kabala: The code change as we talked

pkg/prometheus/prometheus.go
moshe-kabala: I removed the parallel function

moshe-kabala:
moved to https://github.com/run-ai/runai-cli/pull/30

PR #30, opened 2020-10-12
RUN-967 - top node command

moshe-kabala:


moshe-kabala:
move to https://github.com/run-ai/runai-cli/pull/32

PR #37, opened 2020-10-14
Ability to resolve conflict, applied only on delete

omer-dayan:


pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+func generateConflictError(conflictedJobs []trainer.TrainingJob) error {
+	message := fmt.Sprintf("There are more than one training job with the name %s: \n", conflictedJobs[0].Name())
+	for i, job := range conflictedJobs {
+		message = fmt.Sprintf("%s \t %d) %s, %s, %s\n", message, i, job.Name(), job.TrainerName(), job.Trainer())
+	}
+	message = fmt.Sprintf("%sTo delete a specifig job you can use the flags --training-type, --interactive, and --train", message)
+	return fmt.Errorf(message)
+}
+
+func guessTrainingJobByTrainer(job JobIdentifier, t trainer.Trainer) ([]trainer.TrainingJob, error) {
```
yodarshafrir1, line 29: `t`-> `trainer`

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+	return matchingJobs[0], nil
+}
+
+func GetTrainingJob(job JobIdentifier, kubeClient *client.Client) (trainer.TrainingJob, error) {
+	var jobTrainer trainer.Trainer
+	var trainingJob trainer.TrainingJob
+
+	var err error
+	switch job.Trainer {
+	case "runaijob":
```
yodarshafrir1, line 73: Please add a global const with this value

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+		jobTrainer = trainer.NewRunaiTrainer(*kubeClient)
+	case "mpijob":
```
yodarshafrir1, line 75: Please add a global const with this value

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+	"github.com/run-ai/runai-cli/cmd/trainer"
+	"github.com/run-ai/runai-cli/pkg/client"
+)
+
+const (InteractiveJobTrainerLabel = "Interactive"
+		TrainJobTrainerLabel = "Train")
+
+type JobIdentifier struct {
```
yodarshafrir1, line 12: The convention here is to name in snake case.

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+	trainingJobs, err := jobTrainer.GetTrainingJobs(job.Name, job.Namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(trainingJobs) == 1 {
+		return trainingJobs[0], nil
+	} else if job.Interactive {
+		for _, trainingJob := range trainingJobs {
+			if (trainingJob.Trainer() == "Interactive") == job.Interactive {
```
yodarshafrir1, line 94: Please consider renaming `Trainer` not a must, but it might be confusing with `trainer.Trainer`

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+type JobIdentifier struct {
+	Name          string
+	Namespace     string
+	Trainer       string
+	Interactive   bool
+	Train		  bool
+}
+
+func generateConflictError(conflictedJobs []trainer.TrainingJob) error {
+	message := fmt.Sprintf("There are more than one training job with the name %s: \n", conflictedJobs[0].Name())
```
yodarshafrir1, line 21: `training` can be confusing for the users since we use that string for both train and interactive

cmd/job/delete/delete.go
```diff
@@ -59,63 +55,37 @@ func NewDeleteCommand() *cobra.Command {
...
 				if err != nil {
 					log.Error(err)
 				}
 			}
 		},
 	}
 
+	command.Flags().BoolVarP(&typeInteractive, "interactive", "", false, "Specifies whether to delete interactive job")
+	command.Flags().BoolVarP(&typeTrain, "train", "", false, "Specifies whether to delete train job")
+	command.Flags().StringVarP(&trainerType, "training-type", "", "", "Specifies the trainer type to avoid conflict")
```
yodarshafrir1, line 69: I think `training-type` can be confusing, because we have train and interacitve

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+	for _, trainer := range trainers {
+		trainingJobs, err := guessTrainingJobByTrainer(job, trainer)
+		if err != nil {
+			continue
+		}
+		matchingJobs = append(matchingJobs, trainingJobs...)
+	}
+
+	if len(matchingJobs) == 0 {
+		return nil, fmt.Errorf("there is not job name %s in namespace %s", job.Name, job.Namespace)
```
yodarshafrir1, line 60: The users aren't familiar with namespace, you should use the `project` in your message, please see the message we have today

pkg/jobs/jobFinder.go
```diff
@@ -0,0 +1,108 @@
...
+
+type JobIdentifier struct {
+	Name          string
+	Namespace     string
+	Trainer       string
+	Interactive   bool
+	Train		  bool
+}
+
+func generateConflictError(conflictedJobs []trainer.TrainingJob) error {
```
yodarshafrir1, line 20: Can we set a specific message for every case.
If 2 trainingtypes exist with add the flag training-type
If 1 training and we have interactive and train add interactive flag
if 2 above apply add both flags

pkg/jobs/jobFinder.go
yodarshafrir1, line 94: We also have a job `preemptible-interactive`
we want the interactive flag to delete this job too ( this job is a statefulset) please verify it is supported.

pkg/util/kubectl/kubectl.go
```diff
@@ -361,7 +360,7 @@ func SaveAppConfigMapToFile(name, key, namespace string) (fileName string, err e
 	args = append(args, ">", fileName)
 
 	log.Debugf("Exec bash -c %s", strings.Join(args, " "))
-
+	fmt.Println(args)
```
yodarshafrir1, line 363: please delete?

pkg/workflow/workflow.go
```diff
@@ -19,33 +21,26 @@ import (
...
-	}
-
-	_, err = clientset.CoreV1().ConfigMaps(namespace).Get(configMapName, metav1.GetOptions{})
-
-	if err != nil {
-		log.Debugf("Skip deletion of ConfigMap %s, because the ConfigMap does not exist.", configMapName)
-		return nil
-	}
+func DeleteJob(job trainer.TrainingJob, kubeClient *client.Client) error {
+	go (func() {
```
yodarshafrir1, line 25: Why use goroutine?

pkg/jobs/jobFinder_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"github.com/magiconair/properties/assert"
+	"github.com/run-ai/runai-cli/cmd/trainer"
+	cmdTypes "github.com/run-ai/runai-cli/pkg/types"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+type trainerMock struct {
```
yodarshafrir1, line 12: 🥇

cmd/job/delete/delete.go
```diff
@@ -59,63 +55,37 @@ func NewDeleteCommand() *cobra.Command {
...
+				err = DeleteJob(maybeJobIdentifier, kubeClient)
 				if err != nil {
 					log.Error(err)
 				}
 			}
 		},
 	}
 
+	command.Flags().BoolVarP(&typeInteractive, "interactive", "", false, "Specifies whether to delete interactive job")
+	command.Flags().BoolVarP(&typeTrain, "train", "", false, "Specifies whether to delete train job")
```
yodarshafrir1, line 68: Until now for `train` jobs we didn't ask the user to add any flags 
Please consult with @yarongol whether we want to have 1 flag or 2 flags:
1 flag:
` --is-interactive true` or `--is-interactive false`
2 flags:
`--interactive` or `--train`

pkg/jobs/jobFinder.go
yodarshafrir1: If a user added `interactive` flag and no job was found this message is not right.
There might be a job with this name, but not an interactive job.

pkg/workflow/workflow.go
```diff
@@ -19,33 +21,26 @@ import (
...
-	}
+		err = kubeClient.GetClientset().CoreV1().ConfigMaps(job.Namespace()).Delete(configMapName, &metav1.DeleteOptions{})
+		if err != nil {
+			log.Warningf("Delete configmap %s failed, please clean it manually due to %v.", configMapName, err)
+			log.Warningf("Please run `kubectl delete -n %s cm %s`", job.Namespace(), configMapName)
+		}
+	})()
 
-	return nil
+	return job.Delete(kubeClient)
```
yodarshafrir1, line 43: Currently the job is deleted by the configmap no?
meaning might return an error if the job was deleted by the configmap.

yodarshafrir1 (changes requested):
Looks good, please see my comments

cmd/job/delete/delete.go
yodarshafrir1: Please consult with @yarongol  here as well, users currently are not familiar with the name `training-type`

yodarshafrir1 (changes requested):

cmd/job/delete/delete.go
```diff
@@ -59,63 +55,37 @@ func NewDeleteCommand() *cobra.Command {
 			}
 
 			for _, jobName := range args {
-				err = deleteTrainingJob(kubeClient, jobName, namespaceInfo)
+				maybeJobIdentifier := jobs.JobIdentifier{Name: jobName, Namespace: namespaceInfo.Namespace, Trainer: strings.ToLower(trainerType), Interactive: typeInteractive, Train: typeTrain}
```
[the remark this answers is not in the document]
omer-dayan: "Maybe" prefix is a convention for values that are not meaningful for sure. (Like "is" prefix for boolean). 

you can read here:
https://en.wikipedia.org/wiki/Option_type

"it is used as the return type of functions which may or may not return a meaningful value when they are applied"

pkg/jobs/jobFinder.go
omer-dayan: if I change the argument name it hides the trainer namespace so I can no longer able to use it
maybe "chosenTrainer"? that sound pretty weird to me

pkg/jobs/jobFinder.go
omer-dayan: All the strings I will go throught with Yaron

pkg/workflow/workflow.go
omer-dayan: it wasnt go routine at the start. and then I noticed we delete the job itself no matter what happens in to the resources. so why wait with that

PR #39, opened 2020-10-18
RUN-1110 - Submit a job with unique name across all the "job types" namespace and generate suffix

omer-dayan:


pkg/workflow/workflow.go
```diff
@@ -127,172 +118,166 @@ func generateJobFiles(name string, namespace string, values interface{}, environ
...
-		if err != nil {
-			log.Warnf("Failed to delete %s due to %v", jobFiles.appInfoFileName, err)
+		configMap, err = createEmptyConfigMap(maybeConfigMapName, name, namespace, configMapIndex, clientset)
+		if err == nil {
+			return configMap, nil
 		}
 	}
 
-	return nil
+	return nil, fmt.Errorf("could not create job, please try again later")
```
yodarshafrir1, line 174: How about?
`Job creation has failed. Please try again`

pkg/workflow/workflow.go
```diff
@@ -127,172 +118,166 @@ func generateJobFiles(name string, namespace string, values interface{}, environ
...
-	_, err = clientset.CoreV1().ConfigMaps(namespace).Create(&configMap)
-	return err
+func submitJobInternal(name, namespace string, generateName bool, values interface{}, environmentValues string, chart string, clientset kubernetes.Interface) (string, error) {
+	configMap, err := submitConfigMap(name, namespace, generateName, clientset)
+	if err != nil {
+		return "", err
+	}
+	jobName := configMap.Name
+	jobFiles, err := generateJobFiles(jobName, namespace, values, environmentValues, chart)
+	if err != nil {
```
yodarshafrir1, line 245: Don't we first want to call the `defer` ?

cmd/job/list/list.go
```diff
@@ -119,5 +142,9 @@ func displayTrainingJobList(jobInfoList []trainer.TrainingJob, displayGPU bool)
 			runningOfActivePods,
 			strings.Join(jobInfo.ServiceURLs(), ", "))
 	}
+
+	for _, invalidJob := range invalidJobs {
+		ui.Line(w, invalidJob, "Invalid job. Please delete it!", "", "", "", "", "", "", "", "", "")
```
yodarshafrir1, line 147: Since you put so much text in the `STATUS` column it can cause an indentation issue

cmd/job/submit/submit.go
```diff
@@ -206,6 +207,7 @@ func (submitArgs *submitArgs) addCommonFlags(fbg flags.FlagsByGroups) {
 	// Will not submit the job to the cluster, just print the template to the screen
 	flagSet.BoolVar(&dryRun, "dry-run", false, "Run as dry run")
 	flagSet.MarkHidden("dry-run")
+	flags.AddBoolNullableFlag(flagSet, &submitArgs.GenerateName, "generate-name", "", "Generate suffix to the job if the original name is occupied")
```
yodarshafrir1, line 210: `occupied` -> `used`

yodarshafrir1 (changes requested):

pkg/workflow/workflow.go
```diff
@@ -15,34 +15,40 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	FamilyNameLabelSelectorName = "FamilyName"
```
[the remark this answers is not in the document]
omer-dayan: Except of FamilyNameLabelSelectorName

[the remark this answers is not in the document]
omer-dayan: Lack of better term to describe the connection among all those that has the same base name

pkg/workflow/workflow.go
omer-dayan: If there is an error it means the jobFiles object is not generated. no need to clean it in such case

cmd/job/list/list.go
omer-dayan: The width of the column derived from the widest cell in that column. This text seems pretty fine compared to other text can be in this column
Screen Shot 2020-10-19 at 15 25 15

omer-dayan: Its seems natural (In my opinion) next to the ImagePullBackOff

pkg/workflow/workflow.go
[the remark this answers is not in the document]
yodarshafrir1: Isn't just ruins from the part you created different prefix for different family, 1 family was the MPI and 1 was RunaiJob?
Why not just baseNameLabel or baseLabelSelector? we no longer have anything to do with families

pkg/workflow/workflow.go
yodarshafrir1: If `generateJobFiles` returns an error at the last part of the method, on `kubectl.SaveAppInfo(template, namespace)` we will have files, don't we?

cmd/job/list/list.go
```diff
@@ -71,9 +73,30 @@ func NewListCommand() *cobra.Command {
...
+				jobsMap[job.Name()] = true
+			}
+
+			configMaps, err := kubeClient.GetClientset().CoreV1().ConfigMaps(namespaceInfo.Namespace).List(metav1.ListOptions{})
+			if err != nil {
+				log.Errorf("Failed due to %v", err)
+				os.Exit(1)
+			} else {
+				for _, item := range configMaps.Items {
+					if item.Labels[workflow.FamilyNameLabelSelectorName] != "" {
```
yodarshafrir1, line 88: Not sure why do we need this condition?
Why not always show the row on the job list if configmap exist and job doesn't?
Also every new job will have this label, and we want to present old configmaps as well.
Am I missing something?

cmd/job/list/list.go
yodarshafrir1: The `list` command is very wide already, we don't want to make it wider and we don't want to tell the user what actions to do on the status.
You can show the status as: `StaleJob` or `Stale` or `InvalidJob` or `Invalid`
If the user will want to get more information he can run the `get` command and understand that the job should be deleted.

yodarshafrir1 (changes requested):
Please also see my notes on the previous review

cmd/job/list/list.go
omer-dayan: I Definitaly agree about that!

cmd/get/get.go
```diff
@@ -206,6 +205,10 @@ func getTrainingJobsByName(kubeClient *client.Client, name string, namespaceInfo
 
 	if len(jobs) == 0 {
 		log.Debugf("Failed to find the training job %s in namespace %s", name, namespaceInfo.Namespace)
+		configMap, err := kubeClient.GetClientset().CoreV1().ConfigMaps(namespaceInfo.Namespace).Get(name, metav1.GetOptions{})
```
yodarshafrir1, line 208: What about all the other commands (bash, exec, logs)?

yodarshafrir1 (changes requested):

pkg/workflow/workflow.go
omer-dayan: Yes, but in case of error the object return nil. And its better to handle this inside the generateJobFiles (Done)

cmd/get/get.go
omer-dayan: Unfortunately all of the other commands use the function in the GET package. that means all the commands gets to this line of code

pkg/workflow/workflow.go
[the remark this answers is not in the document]
omer-dayan: BaseNameLabelSelector is chosen! I didnt like the family work either

cmd/job/list/list.go
omer-dayan: Thats the exact reason I did it. From now on, we dont do config map name like "test-runaijob" or "test-mpijob". if the job called "test" then the configmap is called "test". So old cases need seperate logic (However, I dont think its worth doing - its a feature we add to the CLI, it make sense its not working retroactive)

pkg/workflow/workflow.go
```diff
@@ -15,34 +17,46 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	BaseNameLabelSelectorName  = "FamilyName"
```
yodarshafrir1, line 21: `FamilyName` -> `base-name`?

pkg/workflow/workflow.go
```diff
@@ -15,34 +17,46 @@ import (
...
+	baseIndexLabelSelectorName = "FamilyIndex"
```
yodarshafrir1, line 22: `FamilyIndex` ->`base-name-index` ?

pkg/workflow/workflow.go
yodarshafrir1: Great

pkg/workflow/workflow.go
```diff
@@ -127,172 +128,168 @@ func generateJobFiles(name string, namespace string, values interface{}, environ
...
-			log.Debugf("Failed to uninstall app with configmap.")
-		}
-		cleanUpErr = kubectl.DeleteAppConfigMap(configMapName, namespace)
-		if cleanUpErr != nil {
-			log.Debugf("Failed to cleanup configmap %s", configMapName)
-		}
-
-		return fmt.Errorf("Failed submitting the job:\n %s", err.Error())
+	if !generateName {
+		return nil, fmt.Errorf("there is another job with the name %s, you can either delete it or you can use the --generate-name flag", maybeConfigMapName)
```
yodarshafrir1, line 166: this is the old message: 
`The job <job-name> already exists, please delete it first. use 'runai delete <job-name>`
so maybe use this string:
`The job %s already exists, either delete it first (use 'runai delete <job-name>' ) or submit the job again using the flag  --generate-name`

pkg/workflow/workflow.go
```diff
@@ -127,172 +128,168 @@ func generateJobFiles(name string, namespace string, values interface{}, environ
...
+	return jobName, nil
+}
+
+func SubmitJob(name, namespace string, generateName bool, values interface{}, environmentValues string, chart string, clientset kubernetes.Interface, dryRun bool) (string, error) {
+	if dryRun {
+		jobFiles, err := generateJobFiles(name, namespace, values, environmentValues, chart)
+		if err != nil {
+			return "", err
+		}
+		fmt.Println("Generate the template on:")
```
yodarshafrir1, line 286: do we really want this print?

pkg/workflow/workflow.go
```diff
@@ -127,172 +128,168 @@ func generateJobFiles(name string, namespace string, values interface{}, environ
...
+		fmt.Println(jobFiles.template)
```
yodarshafrir1, line 287: do we really want this print?

pkg/workflow/workflow.go
```diff
@@ -127,172 +128,168 @@ func generateJobFiles(name string, namespace string, values interface{}, environ
...
+		return jobName, err
+	}
+
+	err = populateConfigMap(configMap, chartName, chartVersion, jobFiles.envValuesFile, jobFiles.valueFileName, jobFiles.appInfoFileName, namespace, clientset)
+	if err != nil {
+		return jobName, err
+	}
+
+	_, err = kubectl.InstallApps(jobFiles.template, namespace)
+	fmt.Println(jobFiles.template)
```
yodarshafrir1, line 273: do we really want this print?

cmd/job/list/list.go
```diff
@@ -71,9 +73,30 @@ func NewListCommand() *cobra.Command {
...
+			configMaps, err := kubeClient.GetClientset().CoreV1().ConfigMaps(namespaceInfo.Namespace).List(metav1.ListOptions{})
+			if err != nil {
+				log.Errorf("Failed due to %v", err)
+				os.Exit(1)
+			} else {
+				for _, item := range configMaps.Items {
+					if item.Labels[workflow.BaseNameLabelSelectorName] != "" {
+						if jobsMap[item.Name] == false {
+							fmt.Println(item.Name)
+							invalidJobs = append(invalidJobs, item.Name)
```
yodarshafrir1, line 91: do we really want this print?

cmd/job/list/list.go
```diff
@@ -71,9 +73,30 @@ func NewListCommand() *cobra.Command {
...
+			}
+
+			configMaps, err := kubeClient.GetClientset().CoreV1().ConfigMaps(namespaceInfo.Namespace).List(metav1.ListOptions{})
+			if err != nil {
+				log.Errorf("Failed due to %v", err)
+				os.Exit(1)
+			} else {
+				for _, item := range configMaps.Items {
+					if item.Labels[workflow.BaseNameLabelSelectorName] != "" {
+						if jobsMap[item.Name] == false {
```
yodarshafrir1, line 89: let's discuss this

yodarshafrir1 (changes requested):
last cycle

cmd/job/delete/delete.go
```diff
@@ -59,7 +50,7 @@ func NewDeleteCommand() *cobra.Command {
 			}
 
 			for _, jobName := range args {
-				err = deleteTrainingJob(kubeClient, jobName, namespaceInfo)
+				err = workflow.DeleteJob(jobName, namespaceInfo, kubeClient.GetClientset())
```
yodarshafrir1, line 53: Please add backward compatibility for deletion of old jobs

yodarshafrir1 (changes requested):

pkg/workflow/workflow.go
omer-dayan: Its a dry-run flag. We use to print it in the past.
You add dry run and get the YAML that is going to be sent to the server

pkg/workflow/workflow.go
omer-dayan: Its a dry-run flag. We use to print it in the past.
You add dry run and get the YAML that is going to be sent to the server

yodarshafrir1 (changes requested):
Good job!
Let's wait with pushing this for the 2nd PR

yodarshafrir1 (approved):
🥇

PR #41, opened 2020-10-21
Alignment to what has been chosen

omer-dayan:


cmd/job/submit/submit.go
```diff
@@ -258,16 +259,25 @@ func (submitArgs *submitArgs) addCommonFlags(fbg flags.FlagsByGroups) {
 func (submitArgs *submitArgs) setCommonRun(cmd *cobra.Command, args []string, kubeClient *client.Client, clientset kubernetes.Interface, configValues *string) error {
 	util.SetLogLevel(global.LogLevel)
 	var name string
-	if nameParameter == "" && len(args) >= 1 {
+	if nameParameter != "" {
+		if len(args) > 0 || submitArgs.namePrefix != "" {
+			submitArgs.generateSuffix = false
```
yodarshafrir1, line 264: Isn't this the default value anyway?

cmd/job/submit/submit.go
```diff
@@ -258,16 +259,25 @@ func (submitArgs *submitArgs) addCommonFlags(fbg flags.FlagsByGroups) {
...
+			log.Warn("Using the name provided with --name flag. Ignoring the other options")
+		}
+		name = nameParameter
+	} else if len(args) > 0 {
+		if submitArgs.namePrefix != "" {
+			log.Warn("Ignoring --job-name-prefix flag")
+			submitArgs.generateSuffix = false
+		}
+		log.Warn("Provide name without --name flag has been deprecated")
```
yodarshafrir1, line 273: cobra has a message for deprecated flags, i just ran: `runai submit --name blah`
`Flag --name has been deprecated, please use positional argument instead`
So I guess this is better:
`Submitting the job name as a positional argument has been deprecated, please use --name flag instead`.
 Also let's not use warn here?

cmd/job/submit/submit_runai.go
```diff
@@ -29,16 +29,16 @@ import (
 const (
 	submitExamples = `
 # Start a Training job.
-runai submit train1 -i gcr.io/run-ai-demo/quickstart -g 1
+runai submit --name train1 -i gcr.io/run-ai-demo/quickstart -g 1
```
yodarshafrir1, line 32: 🥇

pkg/workflow/workflow.go
```diff
@@ -160,34 +160,33 @@ func getSmallestUnoccupiedIndex(configMaps []corev1.ConfigMap) int {
 	return len(configMaps)
 }
 
-func getConfigMapName(name string, index int) string {
-	if index == 0 {
+func getConfigMapName(name string, index int, generateSuffix bool) string {
+	if !generateSuffix {
```
yodarshafrir1, line 164: Didn't we say we want to start from index 0?

cmd/job/submit/submit.go
```diff
@@ -258,16 +259,25 @@ func (submitArgs *submitArgs) addCommonFlags(fbg flags.FlagsByGroups) {
 func (submitArgs *submitArgs) setCommonRun(cmd *cobra.Command, args []string, kubeClient *client.Client, clientset kubernetes.Interface, configValues *string) error {
 	util.SetLogLevel(global.LogLevel)
 	var name string
-	if nameParameter == "" && len(args) >= 1 {
+	if nameParameter != "" {
+		if len(args) > 0 || submitArgs.namePrefix != "" {
+			submitArgs.generateSuffix = false
+			log.Warn("Using the name provided with --name flag. Ignoring the other options")
```
yodarshafrir1, line 265: Let's rephrase this

yodarshafrir1 (changes requested):
Please see my notes

cmd/job/submit/submit.go
omer-dayan: You sure right

pkg/workflow/workflow.go
omer-dayan: Yes as you can see in the line 171 we call the method the first attempt with index 0

yodarshafrir1 (changes requested):

yodarshafrir1 (approved):
💯

PR #45, opened 2020-10-22
RUN-1129 adding createBy=RunaiJob label in the runai trainer

moshe-kabala:


cmd/trainer/trainer_runai.go
```diff
@@ -202,6 +204,10 @@ func (rt *RunaiTrainer) getRunaiTrainingJob(podSpecJob cmdTypes.PodTemplateJob,
 	}
 
 	labels := []string{}
+	// add created by runai label if it don't existed
```
yodarshafrir1, line 207: You should filter the `get` command by this label.
Why add this label?
What about the bug you needed to fix where the CLI crashed?

yodarshafrir1 (changes requested):

cmd/trainer/trainer_runai.go
moshe-kabala: The get action will be filtered according to these labels
So the problem is solved because I only got the pods created by runai.
I added another check (pod.OwnerReferences! = Zero) but this case should not happen

moshe-kabala: Also, I already test it to see that the problem solved

yodarshafrir1: We want to show all jobs/pods, not only those who were created by the cli.
This is not the right solution

moshe-kabala: I assume from the code that the TrainRunai get the runai job and latter search for its pods and these pod should to contain the createBy: runai label. 
where I am wrong ?

moshe-kabala:
just adding the `pod.OwnerReferences != nil`  to master

PR #53, opened 2020-10-29
Ability to make field mandatory - template stage 2

omer-dayan:


cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
+
+func applyTemplateFieldForFloat64(cliFlag *float64, templateField *templates.TemplateField, fieldName string) *float64 {
+	var value *float64
+	required := false
+	if templateField != nil {
+		required = raUtil.IsBoolPTrue(templateField.Required)
+		templateFieldValue, err := strconv.ParseFloat(templateField.Value, 64)
+		if err != nil && templateField.Value != "" {
+			value = mergeFloat64Flags(cliFlag, &templateFieldValue)
+		} else {
```
yodarshafrir1, line 193: can we merge this with the next lines?

cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
+
+func applyTemplateFieldForInt(cliFlag *int, templateField *templates.TemplateField, fieldName string) *int {
+	var value *int
+	required := false
+	if templateField != nil {
+		required = raUtil.IsBoolPTrue(templateField.Required)
+		templateFieldValue, err := strconv.Atoi(templateField.Value)
+		if err != nil && templateField.Value != "" {
+			value = mergeIntFlags(cliFlag, &templateFieldValue)
+		} else {
```
yodarshafrir1, line 214: same

cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
+
+func applyTemplateFieldForBool(cliFlag *bool, templateField *templates.TemplateField, fieldName string) *bool {
+	var value *bool
+	required := false
+	if templateField != nil {
+		required = raUtil.IsBoolPTrue(templateField.Required)
+		templateFieldValue, err := strconv.ParseBool(templateField.Value)
+		if err != nil && templateField.Value != "" {
+			value = mergeBoolFlags(cliFlag, &templateFieldValue)
+		} else {
```
yodarshafrir1, line 235: same

cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
+		if err != nil && templateField.Value != "" {
+			value = mergeDurationFlags(cliFlag, &templateFieldValue)
+		} else {
+			value = mergeDurationFlags(cliFlag, nil)
+		}
+	} else {
+		value = mergeDurationFlags(cliFlag, nil)
+	}
+
+	if value == nil && required {
```
yodarshafrir1, line 263: Can you create a method to check if value is nil and required and panics?

yodarshafrir1 (changes requested):

cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
+	return value
+}
+
+func applyTemplateFieldForBool(cliFlag *bool, templateField *templates.TemplateField, fieldName string) *bool {
+	var value *bool
+	required := false
+	if templateField != nil {
+		required = raUtil.IsBoolPTrue(templateField.Required)
+		templateFieldValue, err := strconv.ParseBool(templateField.Value)
+		if err != nil && templateField.Value != "" {
```
yodarshafrir1, line 233: if error print that we failed to take value from tempalte?

yodarshafrir1 (changes requested):

cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
+}
+
+
+func applyTemplateFieldForFloat64(cliFlag *float64, templateField *templates.TemplateField, fieldName string) *float64 {
+	var value *float64
+	required := false
+	if templateField != nil {
+		required = raUtil.IsBoolPTrue(templateField.Required)
+		templateFieldValue, err := strconv.ParseFloat(templateField.Value, 64)
+		if err != nil && templateField.Value != "" {
```
yodarshafrir1, line 191: Doesn't the cli need to override this value?

yodarshafrir1 (changes requested):

cmd/job/submit/submit_template_util.go
```diff
@@ -1,13 +1,21 @@
...
 	raUtil "github.com/run-ai/runai-cli/cmd/util"
 	"github.com/run-ai/runai-cli/pkg/templates"
+	"strconv"
 	"strings"
 	"time"
 )
 
-func applyTemplateToSubmitRunaijob(templateYaml string, args *submitRunaiJobArgs, extraArgs []string) error {
+func applyTemplateToSubmitRunaijob(templateYaml string, args *submitRunaiJobArgs, extraArgs []string) (err error) {
+	defer (func() {
```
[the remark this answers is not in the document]
omer-dayan: Its the function that translate a panic to an error in case of missing flag.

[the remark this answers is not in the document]
omer-dayan: Moved to different function

cmd/job/submit/submit_template_util.go
```diff
@@ -156,12 +171,113 @@ func mergeExtraArgs(cliExtraArgs, templateExtraArgs []string) []string {
...
 		submitArgs.SpecCommand = []string{}
 		submitArgs.SpecArgs = mergeExtraArgs(extraArgs, template.ExtraArgs)
 	}
+}
+
+
+func applyTemplateFieldForFloat64(cliFlag *float64, templateField *templates.TemplateField, fieldName string) *float64 {
+	var value *float64
+	required := false
+	if templateField != nil {
```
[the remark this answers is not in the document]
omer-dayan: Yes, the default template is the "templateField" in this case and its changing the required value

cmd/job/submit/submit_template_util.go
omer-dayan: Yes it does

cmd/job/submit/submit_template_util.go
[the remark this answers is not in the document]
yodarshafrir1: Moshe is currently working on setting some fields as required by cobra.
Before your PR he could have used https://github.com/spf13/cobra/issues/206#issuecomment-471959800 `cobra.MarkFlagRequired()`

We can set a flag that will be required in every run, if they will not be mentioned in the template we want to fail the job.
Can you think about this case? what would be the right solution? check if the flag is required by cobra first? not use cobra to set a required fields and add checks after the templates logic?

cmd/job/submit/submit_template_util.go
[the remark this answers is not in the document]
yodarshafrir1: looks great!

cmd/job/submit/submit_template_util.go
```diff
@@ -17,7 +28,13 @@ func applyTemplateToSubmitRunaijob(templateYaml string, args *submitRunaiJobArgs
 	return nil
 }
 
-func applyTemplateToSubmitMpijob(templateYaml string, args *submitMPIJobArgs, extraArgs []string) error {
+func applyTemplateToSubmitMpijob(templateYaml string, args *submitMPIJobArgs, extraArgs []string) (err error) {
+	defer (func() {
+		if r := recover(); r != nil {
```
yodarshafrir1, line 33: Can you use the method  `recoverFromMissingFlag` here too

yodarshafrir1 (changes requested):

PR #72, opened 2020-11-26
Deletion job resources even in case configmap not found

omer-dayan:


pkg/workflow/workflow.go
```diff
@@ -64,16 +67,81 @@ func DeleteJob(jobName string, namespaceInfo types.NamespaceInfo, clientset kube
...
+		log.Warningf("Delete configmap %s failed due to %v. Please clean it manually", configMapName, err)
 		log.Warningf("Please run `kubectl delete -n %s cm %s`", namespace, configMapName)
 		return err
 	}
 
 	return nil
 }
 
+func deleteJobResourcesWithoutConfigMap(jobName string, namespaceInfo types.NamespaceInfo, clientset kubernetes.Interface) error {
+	client, err := client.GetClient()
```
yodarshafrir1, line 81: can't we use clientset?

pkg/workflow/workflow.go
```diff
@@ -64,16 +67,81 @@ func DeleteJob(jobName string, namespaceInfo types.NamespaceInfo, clientset kube
...
+	if err != nil {
+		return err
+	}
+
+	matchedJobs, err := trainer.GetTrainingJobsByTypeMap(jobName, namespaceInfo.Namespace, client)
```
yodarshafrir1, line 86: cant you use `SearchTrainingJob`

pkg/workflow/workflow.go
```diff
@@ -64,16 +67,81 @@ func DeleteJob(jobName string, namespaceInfo types.NamespaceInfo, clientset kube
...
+	if err != nil {
+		return err
+	} else if len(matchedJobs) == 0 {
+		return cmdUtil.GetJobDoesNotExistsInNamespaceError(jobName, namespaceInfo)
+	} else if len(matchedJobs) > 1 {
```
yodarshafrir1, line 91: nice

pkg/workflow/workflow.go
```diff
@@ -64,16 +67,81 @@ func DeleteJob(jobName string, namespaceInfo types.NamespaceInfo, clientset kube
...
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func deleteInteractiveJobResources(jobName, namespace string, clientset kubernetes.Interface) {
+	if _, err := clientset.AppsV1().StatefulSets(namespace).Get(jobName, metav1.GetOptions{}); err == nil {
```
yodarshafrir1, line 125: why use get first? why not just delete?

pkg/workflow/workflow.go
```diff
@@ -64,16 +67,81 @@ func DeleteJob(jobName string, namespaceInfo types.NamespaceInfo, clientset kube
...
+		err := clientset.AppsV1().StatefulSets(namespace).Delete(jobName, &metav1.DeleteOptions{})
+		if err != nil {
+			log.Warnf(fmt.Sprintf("Failed to remove statefulSet %v, it may be removed manually and not by using Run:AI CLI.", jobName))
+		}
+	}
+	if _, err := clientset.CoreV1().Services(namespace).Get(jobName, metav1.GetOptions{}); err == nil {
```
yodarshafrir1, line 131: do we really need to delete the services and the ingress?

yodarshafrir1 (changes requested):
please change this to debug as well:
```
if err != nil {
			log.Warnf("Failed to remove some of the job's resources, they might have been removed manually and not by using Run:AI CLI.")
		}

```

pkg/workflow/workflow.go
omer-dayan: In case the job was interactive we need to delete it, why not?

pkg/workflow/workflow.go
omer-dayan: `SearchTrainingJob` require to know the map firstly, to say "mpijob" or "runaijob"

pkg/workflow/workflow.go
omer-dayan: clientset have an access to the basic k8s object like Statefulset/ ingress/ pod/ etc.
For being able to access to runaijob/ mpijob we need the extension to clientset, which build with client.

But for interactiveJobs we use the clientset

yodarshafrir1: please move this to mpi and runai then

pkg/workflow/workflow.go
yodarshafrir1: please check

PR #77, opened 2020-12-03
RUN-1233 - Assign logged in user to jobs where possible

85danf:


cmd/job/submit/submit.go
```diff
@@ -292,6 +287,19 @@ func (submitArgs *submitArgs) setCommonRun(cmd *cobra.Command, args []string, ku
...
 
+func assignUser(submitArgs *submitArgs) {
+	// If its not zero then it was passed as flag by the user
+	if submitArgs.User == "" {
+		if kubeLoginUser, err := auth.GetEmailForCurrentKubeloginToken(); err == nil && kubeLoginUser != "" {
+			// Try to get user from kubelogin cached token
+			submitArgs.User = kubeLoginUser
+		} else if osUser, err := user.Current(); err == nil {
+			// Fallback to OS user
+			submitArgs.User = osUser.Username
```
yodarshafrir1, line 298: why change the old logic and not use:
```
        currentUser, err := user.Current()
	if err != nil {
		defaultUser = ""
	} else {
		defaultUser = currentUser.Username
	}
```

yodarshafrir1: what happens on the first submit, when there's no token?

yodarshafrir1 (changes requested):

cmd/job/submit/submit.go
85danf: How is the new logic 
```
if osUser, err := user.Current(); err == nil {
			// Fallback to OS user
			submitArgs.User = osUser.Username
```

different from the old logic?

```
     currentUser, err := user.Current()
	if err != nil {
		defaultUser = ""
	} else {
		defaultUser = currentUser.Username
	}
```

The difference is purely semantic.


This logic relies on the permission check being run beforehand (thus populating the token as part of the check itself).
In cases where token auth is not used the user will be the OS user, like it was up until now

yodarshafrir1: sorry i though osUser is taken from somewhere else. ignore this comment

yodarshafrir1 (approved):

PR #93, opened 2021-02-03
update-only-if-latest

yarongol:


cmd/update.go
```diff
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/mholt/archiver"
+	arenaVersion "github.com/run-ai/runai-cli/pkg/version"
```
yodarshafrir1, line 15: you mean `runaiVersion` :)

cmd/update.go
```diff
@@ -52,10 +53,20 @@ func NewUpdateCommand() *cobra.Command {
 				os.Exit(1)
 			}
 
+			version, err := arenaVersion.GetVersion()
+			if err != nil {
+				log.Error("Unable to get current cli version")
```
yodarshafrir1, line 58: Can you print the error?

yodarshafrir1 (changes requested):

cmd/update.go
yarongol: done

PR #107, opened 2021-03-17
Adding gpu type

moshe-kabala:


moshe-kabala:
I added a function called SetLabel to define an index label on a string, but I do not use it. I think it might be useful in the feature. If you do not agree let me know and I will remove it

pkg/prometheus/helpers.go
```diff
@@ -20,6 +21,29 @@ func SetFloatFromFirstMetric(num *float64, m MetricResultsByQueryName, key strin
 	return nil
 }
 
+func SetLabel(str *string, label string ,m MetricResultsByQueryName, key string) error {
```
yodarshafrir1, line 24: why do we need this?

pkg/nodes/node_info.go
```diff
@@ -84,6 +84,7 @@ func (ni *NodeInfo) GetResourcesStatus() types.NodeResourcesStatus {
 	nodeResStatus.Allocated = podResStatus.Requested
 	nodeResStatus.Allocated.GPUs = podResStatus.Allocated.GPUs // needed to count fractions as well
 	nodeResStatus.Limited = podResStatus.Limited
+	nodeResStatus.GpuType = ni.Node.Labels["nvidia.com/gpu.product"]
```
yodarshafrir1, line 87: if label not found please set a proper message

yodarshafrir1 (changes requested):
please see my notes and attach a screenshot

pkg/nodes/node_info.go
moshe-kabala: in the field itself? which text? (Not found or - )
by the there is a case of not a GPU node

pkg/helpers/node_resources_status_helpers.go
```diff
@@ -39,6 +39,7 @@ func (c *NodeResourcesStatusConvertor) ToGpus() *types.NodeGPUResource {
 		Usage:       nrs.Usage.GPUs / 100 * float64(capacity),
 		Utilization: nrs.Usage.GPUs,
 		InUse:       nrs.GPUsInUse,
+		Type: 		 nrs.GpuType,
```
yodarshafrir1, line 42: please rename to `GpuType`

pkg/types/node.go
```diff
@@ -16,6 +16,7 @@ type NodeCPUResource struct {
 }
 
 type NodeGPUResource struct {
+	Type        string  `title:"TYPE" def:"-"`
```
yodarshafrir1, line 19: Please rename to `GpuType`

yodarshafrir1 (changes requested):

yodarshafrir1 (approved):

PR #109, opened 2021-04-04
adding runai cli command annotation

moshe-kabala:


charts/mpijob/templates/mpi-job.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
     image: "{{ .Values.image }}"
     totalGPUs: "{{ .Values.totalGpus }}"
     totalGPUsMemory: "{{ .Values.totalGpusMemory }}"
+    runai-cli-command: {{ .Values.runaiCliCommand | quote }}
```
yodarshafrir1, line 13: why did you add it to both mpijob annotation and pod annotation?

charts/runai/templates/runai-job.yaml
```diff
@@ -35,10 +35,11 @@ metadata:
     {{ $key }}: {{ $val | quote }}
     {{- end }}
   {{- end}}
-  {{- if .Values.user }}
   annotations:
+    runai-cli-command: {{ .Values.runaiCliCommand | quote }}
```
yodarshafrir1, line 39: why did you add it to both runaijob annotation and pod annotation?

cmd/job/describe_job.go
```diff
@@ -344,3 +346,10 @@ func getPriorityClass(job trainer.TrainingJob) string {
 
 	return pc
 }
+
+func getCommandLine(job trainer.TrainingJob) ( commandLine string ) {
```
yodarshafrir1, line 350: if no command line write `N/A` or something similiar

cmd/job/submit/submit.go
```diff
@@ -80,6 +83,7 @@ type submitArgs struct {
 	Memory              string   `yaml:"memory,omitempty"`
 	MemoryLimit         string   `yaml:"memoryLimit,omitempty"`
 	EnvironmentVariable []string `yaml:"environment,omitempty"`
+	RunaiCliCommand		string   `yaml:"runaiCliCommand,omitempty"`
```
yodarshafrir1, line 86: `cliCommand`

cmd/job/submit/submit_mpi.go
```diff
@@ -102,6 +102,7 @@ func NewRunaiSubmitMPIJobCommand() *cobra.Command {
 
 	fbg := flags.NewFlagsByGroups(command)
 	submitArgs.addCommonFlags(fbg)
+	submitArgs.addRunaiCliCommand()
```
yodarshafrir1, line 105: you are calling this method from all the job kinds.
better use a common method like: `addCommonSubmit` and from this method call `addCommonFlags` and `addRunaiCliCommand`

cmd/job/submit/submit.go
```diff
@@ -123,6 +127,10 @@ func (s submitArgs) check() error {
 	return nil
 }
 
+func(submitArgs *submitArgs) addRunaiCliCommand() {
```
yodarshafrir1, line 130: why do we need the runai in the name?

yodarshafrir1 (changes requested):

charts/mpijob/templates/mpi-job.yaml
yodarshafrir1: better stick with job's annotation

moshe-kabala: In the scheduler, I extract from the pod but in the researcher UI from the job.

charts/runai/templates/runai-job.yaml
moshe-kabala: In the scheduler, I extract from the pod but in the researcher UI from the job.

cmd/job/submit/submit.go
moshe-kabala: fixed

cmd/job/submit/submit_mpi.go
moshe-kabala: fixed

cmd/job/submit/submit.go
moshe-kabala: fixed

cmd/job/describe_job.go
moshe-kabala: fixed

charts/mpijob/templates/mpi-job.yaml
yodarshafrir1: why not do the same in both?

yodarshafrir1:
Instead of putting the same label on both annotations of job and pods, can we stick to one of them? (the job is better)

charts/mpijob/templates/mpi-job.yaml
moshe-kabala: As agreed I will remain with the 'job' annotation

charts/runai/templates/runai-job.yaml
moshe-kabala: As agreed I will remain with the 'job' annotation

cmd/job/describe_job.go
```diff
@@ -344,3 +346,15 @@ func getPriorityClass(job trainer.TrainingJob) string {
 
 	return pc
 }
+
+func getCliCommand(job trainer.TrainingJob) string {
+	return stringWithDefault(job.CliCommand(), "N/A")
+}
+
+func stringWithDefault(str string, def string) string {
```
omer-dayan, line 354: Its better call "getStringOrDefault", because "with" is not really what we do

omer-dayan: Do we really want seperate function for this?

cmd/job/submit/submit.go
```diff
@@ -80,6 +83,7 @@ type submitArgs struct {
 	Memory              string   `yaml:"memory,omitempty"`
 	MemoryLimit         string   `yaml:"memoryLimit,omitempty"`
 	EnvironmentVariable []string `yaml:"environment,omitempty"`
+	CliCommand		     string  `yaml:"cliCommand,omitempty"`
```
omer-dayan, line 86: Why is it public if its only used within this package?

cmd/job/submit/submit.go
```diff
@@ -123,6 +127,15 @@ func (s submitArgs) check() error {
 	return nil
 }
 
+func(submitArgs *submitArgs) addCommonSubmit(fbg flags.FlagsByGroups) {
+	submitArgs.addCommonFlags(fbg)
+	submitArgs.addCliCommand()
+}
+
+func(submitArgs *submitArgs) addCliCommand() {
+	submitArgs.CliCommand = strings.Join( os.Args , " ")
```
omer-dayan, line 136: NO NO its bad to save it that way, attacker can interfere our YAML file

It should be saved as BASE64

cmd/util/gpu.go
```diff
@@ -32,6 +32,7 @@ const (
 	DeprecatedNVIDIAGPUResourceName    = "alpha.kubernetes.io/nvidia-gpu"
 	PodGroupRequestedGPUs              = "runai-podgroup-requested-gpus"
 	PodGroupRequestedGPUsMemory        = "runai-podgroup-requested-gpus-memory"
+	CliCommand        				   = "runai-cli-command"
```
omer-dayan, line 35: Its not related to GPU, move it to consts.go

omer-dayan (commented):
Liked it,
Please notice its super important to save the CLI Command as BASE64 and not raw text

Also: What about the default template? The CLI command may be mis-leading by its own

cmd/job/describe_job.go
moshe-kabala: deleted

cmd/job/submit/submit.go
moshe-kabala: It private now

cmd/util/gpu.go
moshe-kabala: moved to the const file

cmd/job/submit/submit.go
moshe-kabala: It is OK we are using  quote (in the template)

omer-dayan: Yes its doing escaping but we may lose data when escaping, For example: 

The command : "runai submit -i ubuntu -g 0 --dry-run -- echo \n\t hello"
Becomes : "runai submit -i ubuntu -g 0 --dry-run -- echo nt hello"

omer-dayan (approved):

PR #118, opened 2021-04-19
completion for the runai command line

oferla:


cmd/cluster/list.go
```diff
@@ -29,7 +32,8 @@ func runListCommand(cmd *cobra.Command, args []string) error {
 	for name, context := range config.Contexts {
 		namespace := context.Namespace
 		project := ""
-		if len(namespace) > 6 && namespace[0:6] == "runai-" {
+		lenNsPrefix := len(constants.RUNAI_NS_PROJECT_PREFIX)
```
oferla, line 35: Not exactly related to my fix, but there is a constant for the "runai-" namespace prefix so better use it.

cmd/cluster/list.go
```diff
@@ -29,8 +32,9 @@ func runListCommand(cmd *cobra.Command, args []string) error {
 	for name, context := range config.Contexts {
 		namespace := context.Namespace
 		project := ""
-		if len(namespace) > 6 && namespace[0:6] == "runai-" {
-			project = namespace[6:len(namespace)]
+		lenNsPrefix := len(constants.RUNAI_NS_PROJECT_PREFIX)
+		if len(namespace) > lenNsPrefix && namespace[0:lenNsPrefix] == constants.RUNAI_NS_PROJECT_PREFIX {
+			project = namespace[lenNsPrefix:len(namespace)]
```
oferla, line 37: This change does not exactly related to completion, just noticed there is a constant for "runai-" so better use it...

cmd/constants/const.go
```diff
@@ -2,6 +2,7 @@ package constants
 
 const (
 	RUNAI_QUEUE_LABEL = "runai/queue"
+	RUNAI_NS_PROJECT_PREFIX = "runai-"
```
oferla, line 5: This contact was previously defined in flags.go, which holds flags related constant. Moved it to this place, which is more appropriate.

cmd/flags/const.go
```diff
@@ -2,5 +2,5 @@ package flags
 
 const (
 	ProjectFlag                 = "project"
-	RunaiNamespaceProjectPrefix = "runai-"
```
oferla, line 5: Moved this constant to the const module.

cmd/flags/util.go
```diff
@@ -44,7 +44,11 @@ func getNamespaceToUseFromProjectFlag(cmd *cobra.Command, kubeClient *client.Cli
 	return namespaceInfo, nil
 }
 
-func getNamespaceInfoToUseFromProjectFlag(cmd *cobra.Command, kubeClient *client.Client) (types.NamespaceInfo, error) {
+//    Get namespace from either
+//		- Project flag (-p/--project)
+//		- Default namespace as determined by runai config project command
+//
+func GetNamespaceInfoToUse(cmd *cobra.Command, kubeClient *client.Client) (types.NamespaceInfo, error) {
```
oferla, line 51: Rename the function to more accurate name, as this function does not take only from the flag (there is another function for that) but also from the default namespace.

cmd/job/describe_job.go
```diff
@@ -76,37 +78,43 @@ func DescribeCommand() *cobra.Command {
 				os.Exit(1)
 			}
 			name := args[0]
-
-			kubeClient, err := client.GetClient()
+			job, clientSet, err := PrepareJobInfo(cmd, name)
```
oferla, line 81: Here I just moved the code for finding jobs to a function called PrepareJobInfo, so I can use it from both here and the completion function GenJobNames.

cmd/completion/utils.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+const CACHING_TIME_SEC = 5
```
omer-dayan, line 16: CachingTimeSec

Thats the convention we use.
However there are a lot of places in the CLI where its with '_' but we really want to stick with only one convention

cmd/flags/util.go
omer-dayan: Great

cmd/job/completion.go
```diff
@@ -0,0 +1,142 @@
...
+	//    pods are cahced on a per job basis cause user can change job name while typing thr command
+	//    and in this case we need to re-load the informaiton of the new job
+	//
+	cachePath := COMPLETION_PODS_FILE_SUFFIX + args[0]
+	result := completion.ReadFromCache(cachePath)
+	if result != nil {
+		return result, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	jobInfo, _, err := PrepareJobInfo(cmd, args[0])
```
omer-dayan, line 86: I dont know if you notice, but in the CLI we have a lot of not-efficient code. 
Lots of function that need to only list pods/jobs/nodes/etc and actually do many requests in very not efficient way.
The internal of PrepareJobInfo is one of it.

I wonder if we maybe want to just list all the pods in the namespace - instead of getting all the jobs (There are 4 types of jobs, each is request) and then for each get all the pods. It will be the same in 99% of the times - as users ussually wont run orphan pods in their runai namespace.

It will be much more efficient

cmd/job/completion.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	cachePath := COMPLETION_JOBS_FILE_SUFFIX + "." + namespaceInfo.ProjectName
+
+	result := completion.ReadFromCache(cachePath)
+	if result != nil {
+		return result, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	jobs, invalidJobs, err := PrepareTrainerJobList(kubeClient, namespaceInfo)
```
omer-dayan, line 45: Maybe we want to implement a function that brings only the job names. 

This one makes inside lots of request to know the services, the nodes, ingresses. Data we dont really need, and each request from the API server is really costy

cmd/node/completion.go
```diff
@@ -0,0 +1,21 @@
+package node
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func GenNodeNames(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+
+	nodeInfos, err := GetNodeInfos(false)
```
omer-dayan, line 9: This function do a lot more inside, it gets how many pods are on the node, and lot of thing not really necessery which hurt the performance

maybe just use this :
image

cmd/cluster/list.go
omer-dayan: Lets change it to strings.StartsWith instead of all of this

omer-dayan (commented):
Really loved the feature. And the code too!

Its great that you change things that you find bad and such

Just notice that I mention:
1) Convensions of consts names 
2) Many of the 'GenXNames' use internal functions that are really not efficient, and include lots of unneccesserry API requests, we do not like to use them in many normal cases, especially when the only thing you need is just a list of names. In lots of the cases it would be much more efficient for you to just run request against the Kubernetes API server

omer-dayan:
BTW why won't we install it as part of the Runai CLI installation sh file?

omer-dayan (commented):
Two more things: 
1) There are also 'cmd/root/complete.go' and 'cmd/root/complete_bash.go' - lets remove them.
_**(Ofer) Fixed.**_

2) The cache is saved in a file in /tmp file. There is no way make the cache be in the memory instead of on the FS? As a basic solution I can think of SHM, but Its propaply too much 
_**(Ofer) Fixed.**_

```
        //
 	//   prefer shared memory over /tmp, if shared memory as filesystem exist
 	//
 	tempDir := "/dev/shm"
 	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
 		tempDir = os.TempDir()
 	}

 	return filepath.Join(tempDir, userName + "." + suffix + ".json")
```

omer-dayan (commented):
There are many places where you declare the arguments of function like that:
Screen Shot 2021-04-22 at 9 14 40
If you dont use the arguments/variables in go, you need to change them to '_' like this:
Screen Shot 2021-04-22 at 9 15 43
**_(Ofer) Fixed._**

cmd/completion/utils.go
```diff
@@ -0,0 +1,142 @@
...
+
+//
+//   completion function for commands with no arguments
+//
+func NoArgs(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return nil, cobra.ShellCompDirectiveNoFileComp
+}
+
+func ImagePolicyValues(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return []string { "Always", "IfNotPresent", "Never" }, cobra.ShellCompDirectiveNoFileComp
```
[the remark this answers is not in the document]
oferla: Fixed
func ImagePolicyValues(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	options := []v1.PullPolicy { v1.PullAlways, v1.PullIfNotPresent, v1.PullNever }
	result := make([]string, 0, len(options))
	for _, option := range(options) {
		result = append(result, string(option))
	}
}

cmd/completion/utils.go
oferla: const CachingTimeSec = 5
Fixed.
I was surprised myself to see the non standard constants, after learning that in go the standard is ThisWay, but there was one constant that I encountered which got me confused :-)

cmd/cluster/list.go
oferla: Fixed
```
		project := ""
		if strings.HasPrefix(context.Namespace, constants.RunaiNsProjectPrefix) {
			lenNsPrefix := len(constants.RunaiNsProjectPrefix)
			project = context.Namespace[lenNsPrefix:len(context.Namespace)]
		}
```

cmd/constants/const.go
oferla: Fixed to
```
const (
	RunaiQueueLabel      = "runai/queue"
	RunaiNsProjectPrefix = "runai-"
)
```

cmd/flags/const.go
oferla: Also fixed notation
```
const (
	RunaiQueueLabel      = "runai/queue"
	RunaiNsProjectPrefix = "runai-"
)
```

cmd/node/completion.go
oferla: Fixed.
```
	kubeClient, err := client.GetClient()
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	nodeList, err := kubeClient.GetClientset().CoreV1().Nodes().List(metav1.ListOptions{})
	if err != nil {
		return nil, cobra.ShellCompDirectiveError
	}

	result := make([]string, 0, len(nodeList.Items))

	for _ , nodeItem := range nodeList.Items {
		result = append(result, nodeItem.Name)
	}

```

cmd/job/completion.go
oferla: The plan is to replace this with call to the shared API which we intend to build for the CLI and the researchers UI. Will be considered in this next EPIC.

cmd/job/completion.go
oferla: This will be done as part of next EPIC (see previous comment).

cmd/completion/utils.go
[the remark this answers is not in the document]
omer-dayan: image

Dont you think that the code look simpler like that?

cmd/node/completion.go
omer-dayan: Notice that the List function from kubeClient get metav1.ListOptions{} and you can tell him to bring only the name field and not all the object, it would be more lightweight.

Screen Shot 2021-04-22 at 12 14 21

I am talking about the FieldSelector field

omer-dayan (commented):
About the flag description completion:
Because its not standard in CLI tools - 
Lets have a talk with Eyal, Itay and Yodar and validate that we want it

cmd/completion/utils.go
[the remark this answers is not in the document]
oferla: Did not compile this way first time I did it, but now it works...

omer-dayan (approved):

PR #120, opened 2021-04-20
Adding node status to top node and adding more statuses (v1.NodeConditionType)

moshe-kabala:


cmd/util/node.go
```diff
@@ -44,11 +44,17 @@ func GetNodeInternalAddress(node v1.Node) string {
 	return "unknown"
 }
 
-func IsNodeReady(node v1.Node) bool {
+func GetNodeStatus(node v1.Node) v1.NodeConditionType {
 	for _, condition := range node.Status.Conditions {
-		if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
-			return true
+		if condition.Status == v1.ConditionTrue {
+			return condition.Type
```
omer-dayan, line 50: Why would we limit ourself to a single status?
There can be more that one. For example MemoryPressure and NetworkUnavailable are unrelated and can happen at the same time
Lets show all of them

cmd/util/node.go
```diff
@@ -44,11 +44,17 @@ func GetNodeInternalAddress(node v1.Node) string {
...
 		}
 	}
```
omer-dayan, line 52: Screen Shot 2021-04-20 at 19 07 26

Notice this, we may need to check it either

omer-dayan: (Not sure about that)

cmd/util/node.go
```diff
@@ -44,11 +44,17 @@ func GetNodeInternalAddress(node v1.Node) string {
...
-	return false
+	return v1.NodeConditionType(v1.ConditionUnknown)
+}
+
+func IsNodeReady(node v1.Node) bool {
```
omer-dayan, line 56: Notice please that there is a file named 'tensoboard.go' that use this function. 
However no one use the file anywhere, lets delete it.

omer-dayan, line 56: Screen Shot 2021-04-20 at 19 12 31

Furthermore, this is the only place where we really use this function.
Notice that it rely on the assumption that if node is not in ready state then it has no pods on it. Which is not true in your case.

omer-dayan (commented):
Really good job..

Generally - when you touch the CLI,
תסתכל קצת מסביב לקוד שאתה כותב, יש מלא פונקציות שלא משתמשים בהם, יש המון דברים שכבר לא בשימוש, מלא הערות לא רללונטיות, אל תפחד למחוק

cmd/util/node.go
moshe-kabala: It looks that if the node is not in ready state then it has to **active** pod. Anyway, this is beyond the bug scope.

moshe-kabala: so delete the entire file(tensoboard.go)?

cmd/util/node.go
moshe-kabala: Good to know, I will join them (State1, State2)

cmd/util/node.go
moshe-kabala: I added Unschedulable into the status

omer-dayan (approved):

PR #122, opened 2021-04-22
add validation of /dev/shm

oferla:


cmd/completion/utils.go
```diff
@@ -16,6 +16,8 @@ import (
 
 const CachingTimeSec = 5
 
+const DevShmPath = "/dev/shm"
```
omer-dayan, line 19: const (
        CachingTimeSec = 5
        DevShmPath = "/dev/shm"
)

In go const can be group together

cmd/completion/utils.go
```diff
@@ -141,9 +143,11 @@ func cacheFilePath(suffix string) string {
 	//
 	//   prefer shared memory over /tmp, if shared memory as filesystem exist
 	//
-	tempDir := "/dev/shm"
-	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
-		tempDir = os.TempDir()
+	tempDir := os.TempDir()
+	if _, err := os.Stat(DevShmPath); os.IsExist(err) {
+		if isDirectory, _ := util.IsDirectory(DevShmPath); isDirectory {
```
omer-dayan, line 148: util.IsDirectory is doing os.Stat to DevShmPath. Exacly like the Previous line.
Stat is doing the request to the HardDrive
Why should we do it twice?

it can be changed to: 
if _, err := os.Stat(DevShmPath); os.IsExist(err) && os.IdDir()
Or alternatively put the os.IsExist(err) in the IsDirectory function

oferla: Oh my god this was totally screed up during editing/copying/pasting. The os.Stat in the first line was not meant to be there. Removed, and thanks for noticing!!

cmd/completion/utils.go
oferla: Fixed.
```
  const (
      CachingTimeSec = 5
      DevShmPath = "/dev/shm"
  )
```

omer-dayan (dismissed):
Approved :)

BTW The consts are still seperated

omer-dayan (approved):
:)

PR #125, opened 2021-05-02
remove queues list, we only support projects

oferla:


oferla:
The implementation is incorrect, but the pull request does not resolve it correctly.

PR #126, opened 2021-05-03
Run 1451 inference interactive flags should not work together

natasharomm:


cmd/job/submit/submit_runai.go
```diff
@@ -113,6 +113,10 @@ func NewRunaiJobCommand() *cobra.Command {
 			}
 
 			if raUtil.IsBoolPTrue(submitArgs.Interactive) {
+				if raUtil.IsBoolPTrue(submitArgs.Inference) {
+					fmt.Print("\nThe flags --inference and --interactive cannot be used together\n\n")
```
yodarshafrir1, line 117: Let's use `fmt.Println`

yodarshafrir1 (changes requested):

cmd/job/submit/submit_runai.go
natasharomm: fixed

yodarshafrir1 (approved):

PR #127, opened 2021-05-03
"inference" flag now not hidden

natasharomm:


cmd/job/submit/submit.go
```diff
@@ -65,6 +65,7 @@ type submitArgs struct {
 	Image               string `yaml:"image"`
 	NameParameter       string
 	Project             string `yaml:"project,omitempty"`
+	Inference           *bool  `yaml:"inference,omitempty"`
```
yodarshafrir1, line 68: When adding the flag to this file you actually added this flag to MPI jobs as well. we don't support Inference for MPI jobs. let's move it back to the old file

yodarshafrir1 (changes requested):

cmd/job/submit/submit.go
natasharomm: fixedd

yodarshafrir1 (approved):

PR #128, opened 2021-05-03
Added sorting job list by name then by project

natasharomm:


cmd/trainer/trainer.go
```diff
@@ -78,6 +78,34 @@ func (this orderedTrainingJobByAge) Swap(i, j int) {
 	this[i], this[j] = this[j], this[i]
 }
 
+type orderedTrainingJobByName []TrainingJob
+
+func (this orderedTrainingJobByName) Len() int {
```
yodarshafrir1, line 83: can you name it with a different name rather than `this`, we don't use it in golang

cmd/trainer/trainer.go
```diff
@@ -87,6 +115,24 @@ func MakeTrainingJobOrderdByAge(jobList []TrainingJob) []TrainingJob {
 	return []TrainingJob(newJoblist)
 }
 
+func MakeTrainingJobOrderdByName(jobList []TrainingJob) []TrainingJob {
+	newJoblist := make(orderedTrainingJobByName, 0, len(jobList))
+	for _, v := range jobList {
```
yodarshafrir1, line 120: `job` instead of `v`?

cmd/trainer/trainer.go
```diff
@@ -87,6 +115,24 @@ func MakeTrainingJobOrderdByAge(jobList []TrainingJob) []TrainingJob {
...
+		newJoblist = append(newJoblist, v)
+	}
+	sort.Sort(newJoblist)
+	return []TrainingJob(newJoblist)
+}
+
+func MakeTrainingJobOrderdByProject(jobList []TrainingJob) []TrainingJob {
+	newJoblist := make(orderedTrainingJobByProject	, 0, len(jobList))
```
yodarshafrir1, line 128:
```
var  orderedTrainingJobByProject []TrainingJob
```

cmd/trainer/trainer.go
yodarshafrir1: You created a new object with methods, I'd move it to a different file, it's not related to `trainer.go`

cmd/trainer/trainer.go
```diff
@@ -87,6 +115,24 @@ func MakeTrainingJobOrderdByAge(jobList []TrainingJob) []TrainingJob {
...
+	for _, v := range jobList {
+		newJoblist = append(newJoblist, v)
+	}
+	sort.Sort(newJoblist)
+	return []TrainingJob(newJoblist)
```
yodarshafrir1, line 133: can you return `newJoblist` ? why do we need to create it again?

yodarshafrir1 (changes requested):

cmd/trainer/trainer.go
natasharomm: renamed "this".
I don't agree about the file - there are two other "sorting" functions of training jobs in this file and I wanted to keep them together.

cmd/trainer/trainer.go
natasharomm: fixed

cmd/trainer/trainer.go
natasharomm: var newJoblist orderedTrainingJobByName
(orderedTrainingJobByName is the type)

cmd/trainer/trainer.go
natasharomm: fixed

yodarshafrir1 (approved):

PR #129, opened 2021-05-03
list-projects-via-rs

oferla:


cmd/project/completion.go
```diff
@@ -14,7 +15,12 @@ func GenProjectNamesForFlag(_ *cobra.Command, _ []string, _ string) ([]string, c
 		return result, cobra.ShellCompDirectiveNoFileComp
 	}
 
-	projects, err := PrepareListOfProjects();
+	restConfig, _, err := client.GetRestConfig()
```
oferla, line 18: We may remove the need to work with GetRestConfig once we are premised that we get the URL in the environment variable RESEARCHER_SERVICE_URL. Since this is not completed yet, we also need GetRestConfig for the case that we will have to derive the URL from there.

cmd/project/list.go
```diff
@@ -1,231 +1,141 @@
...
-	projects, err := PrepareListOfProjects();
-	if err != nil {
-		return err
-	}
-
-	// Sort the projects, so they will always appear in the same order
+    //
+    //   obtain default project of this session
+    //
+    restConfig, namespace, err := client.GetRestConfig()
```
oferla, line 30: See my previous comment about RestConfig.

pkg/client/client.go
```diff
@@ -1,13 +1,13 @@
 package client
```
oferla, line 1: This file is only spaces <-> tabs fixing

PR #130, opened 2021-05-05
list projects + delete job to call RS

oferla:
1. Project list + job delete now attempt to call RS REST APIs. 
2. This requires CLI to import code from researcher service GIT repository, as they share the same req/resp struct types, etc. _We're working on resolving the Circle CI issue of accessing private repo._
3. As a temporary solution for upgrade: In case RS is not available, or does not support the necessary API (see versioning mechanism explanation below), the CLI will call the RS function directly, so there will not be duplicate code b/w CLI and RS, in both cases same function runs only different is the way to call it as a REST API or as a function call.
4. RS supports /version API, which returns its build tag. in the CLI, before sending REST API we check that RS supports versioning AND its version is sufficient to handle the API call.
5. In order for CLI to import from RS GIT repo, I had to update go.mod in order to align their packages (there were some conflicts). Specifically, go-client was upgraded to a later version which requires to work with context. Most changes in this pull requests are related to that.
5. Project list now supports --all-deleted flag, in order to show deleted projects (there was a bug which prevented researchers from seeing them). Without the flag we notify the researcher in case he has deleted jobs, and instruct him to use the --all-deleted flag to receive details about them.


cmd/dashboard_helper.go
```diff
@@ -53,7 +54,7 @@ func dashboard(client kubernetes.Interface, namespace string, name string) (stri
 		return url, nil
 	}
 
-	ep, err := client.CoreV1().Endpoints(namespace).Get(name, metav1.GetOptions{})
+	ep, err := client.CoreV1().Endpoints(namespace).Get(context.TODO(), name, metav1.GetOptions{})
```
oferla, line 57: You can ignore those changes, they are related to version upgrade of go-client.

cmd/job/delete/delete.go
```diff
@@ -66,12 +75,63 @@ func NewDeleteCommand() *cobra.Command {
...
+					Project: projectName,
+				})
+			}
+
+			//
+			//    connect to the researcher config, if it can serve delete job request
+			//
+			var deleteJobsStatus []rsrch_server.DeletedJobStatus
+
+			rs := rsrch_client.NewRsrchClient(restConfig, rsrch_client.DeleteJobMinVersion)
```
oferla, line 101: we receive rs object only if RS exist, support versioning, and has a version which supports delete job REST API.

omer-dayan (changes requested):
A few questions about the PR solution (see my answer inline)

1) 
image
What does this files changes related?
### Answer

2) What happens if we have a client like KCL which have older cluster and dont have yet the RS? Do we want to support the old fashion requests? I didnt see you done it in the delete workflow
### Answer 
We nad a dis cussion about it, and the decision was that as for *now*, in case of any failure in contacting the RS, either due to technical issue or because RS has not been upgraded to support the CLI queries, the CLI will invoke the RS method directly. When we'll have significant number of customers that have CLI and RS synced, we may consider to remove the direct calls. When we get to that bridge we'll cross it. For now the flow is REST call -> failure -> direct call. Here is the code:
```
			rs := rsrch_client.NewRsrchClient(restConfig, rsrch_client.DeleteJobMinVersion)
			if rs != nil {
				//
				//   RS can serve the request, so send it to RS
				//
				deleteJobsStatus, err = rs.JobDelete(context.TODO(), jobsToDelete)
			} else {
				log.Infof("researcher-service cannot serve the request, use in-house CLI for job delete")

				clientSet, err := rsrch_cs.NewCliClientFromConfig(restConfig)
				if err != nil {
					log.Errorf("Failed to create clientSet for in-house CLI job delete: %v", err.Error())
					return
				}

				deleteJobsStatus = clientSet.DeleteJobs(context.TODO(), jobsToDelete)
			}
```
The "else" is the direct call, it is invoked if contacting the RS failed for any reason (including versioning mismatch).

3) Do we want to have a RS package in the CLI only so we can have the common objects?
### Answer
Originally, we added the RS package in order for the CLI to reuse the structures and types of RS and to avoid code duplications b/w them. Later on, when the issue of CLI vs old RS came up, we used it also for the "direct call". So instead of leaving all the old code inside CLI, we simply benefit the RS package for direct calling its functions. When/if we decide to remove the "direct call" we will still leave the package for the original purpose.

cmd/job/delete/delete.go
```diff
@@ -66,12 +75,63 @@ func NewDeleteCommand() *cobra.Command {
...
+			} else {
+				for _, deleteJobStatus := range deleteJobsStatus {
+					if deleteJobStatus.Ok {
+						fmt.Printf("Job %s deleted successfully.\n", deleteJobStatus.Name)
+					} else if deleteJobStatus.Error.Status == http.StatusNotFound {
+						fmt.Printf("Job %s does not exist in project %s. If the job exists in a different project, use -p <project-name>.\n", deleteJobStatus.Name, projectName)
+					} else {
+						log.Errorf("%v: %v", deleteJobStatus.Error.Message, deleteJobStatus.Error.Details)
+						fmt.Printf("Failed to delete job %s: %s\n", deleteJobStatus.Name, deleteJobStatus.Error.Message)
+					}
```
omer-dayan, line 131: I am sorry. But I miss something. 

You said that if the RS does not exist then you run the original flow. 
But all I see here that when you run delete command you use only RS flow. Furthermore, you deleted all the logic of deletion from 'workflow.go' (-> submit.go)

cmd/util/util.go
```diff
@@ -55,3 +58,15 @@ func BoolP(b bool) *bool {
...
+func IsProjectNamespace(namespace string) bool {
+	return strings.HasPrefix(namespace, constants.RunaiNsProjectPrefix)
+}
+
+func ToNamespace(project string) string {
+	return constants.RunaiNsProjectPrefix + project
+}
+
+func ToProject(namespace string) string {
+	return namespace[len(constants.RunaiNsProjectPrefix):]
```
omer-dayan, line 71: Just like that? 
Without checking the existence of the 'runai-'? You actually made a really good function for that (IsProjectNamespace)

I am just so sure that people in the future will use this function without checking the input by themself.
"ToProject" its just such a inviting name

cmd/project/completion.go
```diff
@@ -14,7 +15,12 @@ func GenProjectNamesForFlag(_ *cobra.Command, _ []string, _ string) ([]string, c
 		return result, cobra.ShellCompDirectiveNoFileComp
 	}
 
-	projects, err := PrepareListOfProjects();
+	restConfig, _, err := client.GetRestConfig()
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	projects, _, err := PrepareListOfProjects(restConfig, false /*do not include deleted projects*/)
```
omer-dayan, line 23: Also here, someone who does not upgraded the cluster to the version with RS will lose this feature

omer-dayan (commented):
Reviewed almost everything, stoped in the middle because I miss something. Read my comments and tell me what you think

cmd/job/delete/delete.go
oferla: Sent you a demo and will explain to you in the office.

cmd/project/completion.go
oferla: Same here, demo sent explanation in the office :-)

omer-dayan (approved):

PR #133, opened 2021-05-30
RUN-1534

gshaibi:


pkg/nodes/node_info.go
```diff
@@ -86,14 +86,14 @@ func (ni *NodeInfo) GetResourcesStatus() types.NodeResourcesStatus {
 	// check that the totalGpus is set
 	isFractionRunningOnNode := totalGpus > int(nodeResStatus.Capacity.GPUs)
 	if isFractionRunningOnNode {
-		nodeResStatus.NumberOfFractionalAllocatedGpu = len(util.GetSharedGPUsIndexUsedInPods(ni.Pods))
+		nodeResStatus.NumSharedGpus = totalGpus - int(util.GpuCapacity(ni.Node))
```
yodarshafrir1, line 89: before your change we always needed to query the api server for all the running pods on this node - do we still need to do that?

yodarshafrir1 (changes requested):

pkg/nodes/node_info.go
gshaibi: We still calculate other node status properties from the pods.
See several lines above.

yodarshafrir1 (approved):

yodarshafrir1 (approved):

PR #149, opened 2021-07-12
RUN-1630/project namespace independency

oferla:


cmd/cluster/list.go
```diff
@@ -63,12 +63,43 @@ func listCommandDEPRECATED() *cobra.Command {
...
+	if err == nil {
+		core, err := kubernetes.NewForConfig(restConfig)
+		if err == nil {
+			//
+			//   get the labels of the current namespace. if this is a runai namespace, it should
+			//   contain a runai/queue label which holds the name of the project
+			//
+			result, err := core.CoreV1().Namespaces().Get(context.TODO(), kubeContext.Namespace, v1.GetOptions{})
+			if err == nil && result != nil {
+				return result.Labels[constants.RunaiQueueLabel]
```
omer-dayan, line 99: I think its better to use lebel selector instead, instead of request the namespace and then check if the label exists you can do something like that:

result, err := core.CoreV1().Namespaces().Get(context.TODO(), kubeContext.Namespace, v1.GetOptions{LabelSelector: [RunaiQueueLabel]})

Not sure its exactly the syntax

cmd/cluster/list.go
```diff
@@ -30,12 +33,9 @@ func runListCommand(cmd *cobra.Command, args []string) error {
...
 
-	for name, context := range config.Contexts {
-		project := ""
-		if strings.HasPrefix(context.Namespace, constants.RunaiNsProjectPrefix) {
-			lenNsPrefix := len(constants.RunaiNsProjectPrefix)
-			project = context.Namespace[lenNsPrefix:len(context.Namespace)]
-		}
+	for name, kubeContext := range config.Contexts {
+
+		project := getActiveProjectOfCluster(config, name, kubeContext)
```
omer-dayan, line 38: Dont you want to check if the project is empty string? I think its better show "N/A" or just saying that the namespace its not valid Runai Project

omer-dayan (commented):
Looks Really good :)

I didnt see any problem, but let me validate with you:
If old customer will get the CLI with this feature he will still be able to see its projects right?

cmd/cluster/list.go
oferla: This is how the command worked in the past:
```
CLUSTER                                           CURRENT PROJECT
dev-ofer (current)                                gpux2
docker-desktop                                    
gke_run-ai-staging_us-central1-a_staging-cluster  
```

cmd/cluster/list.go
oferla: In label selector you specify set of key-values, the case here is to check that label != "" which is coded using matchExpression. For this case, which I do "get by namespace" and then return it's label from the map there is no added value in complicating the "get" with the weird matchExpression syntax, the rest of the code remains the same and so is execution time. readability of code will be less...

cmd/cluster/list.go
omer-dayan: I dont get it. Dont we want to indicate its a problematic project?

omer-dayan (approved):

cmd/cluster/list.go
oferla: Moved to decision of Itay + Omer, and will open follow up branch if need to apply additional improvement of behavior.

PR #155, opened 2021-07-19
RUN-1735 MIG reports job list

gshaibi:
- Add support for mig-device flag
- List requested MIG instances


cmd/trainer/runai_workload.go
```diff
@@ -266,35 +267,48 @@ func (rj *RunaiWorkload) RequestedGPUMemory() uint64 {
 	return memory
 }
 
-func (rj *RunaiWorkload) RequestedGPUString() string {
-	if memory := rj.RequestedGPUMemory(); memory != 0 {
+func (rw *RunaiWorkload) RequestedMigInstances() []string {
```
gshaibi, line 270: The only real change in this file

cmd/util/gpu.go
```diff
@@ -38,6 +38,7 @@ const (
 	WorkloadCurrentRequestedGPUsMemory = "runai-current-requested-gpus-memory"
 	WorkloadTotalRequestedGPUs         = "runai-total-requested-gpus"
 	WorkloadTotalRequestedGPUsMemory   = "runai-total-requested-gpus-memory"
+	PodAllocatedMigGPUs                = "runai-allocated-mig-gpus"
```
omer-dayan, line 41: In my opinion there is not need to divide it to 2 annotations. One for mig devices and one for fraction GPUs. 
Already mentioned it in the Scheduler CR

cmd/util/mig.go
```diff
@@ -0,0 +1,13 @@
+package util
+
+import "strings"
+
+const migResourcePrefix = "nvidia.com/mig-"
+
+func IsMigInstanceAnnotation(key string) bool {
+	return strings.Contains(key, migResourcePrefix)
```
omer-dayan, line 8: strings.HasPrefix(key, migResourcePrefix)
is more appropriate

pkg/types/node_resources_status.go
```diff
@@ -1,14 +1,14 @@
...
-	GpuType       string
-	Requested     ResourceList
-	Usage         ResourceList
-	NodeGPUs      []GPU
+	Capacity           ResourceList
+	Allocatable        ResourceList
+	Limited            ResourceList
+	Allocated          ResourceList
+	GPUsInUse          int
+	NumSharedOrMigGpus int
```
omer-dayan, line 9: I think that it should stay NumSharedGpus. Mig device is also a method to share GPU

cmd/util/gpu.go
gshaibi: I agree that it would be great if it was united, but there is a big difference between them.
The `fractional-gpus` indicates also the *requested* fractional gpus and is calculated on a submission of a job (appears even if the pod is not bound), while the `runai-allocated-mig-gpus` indicates only *allocated* gpus, and is calculated only upon allocation (prior to that we don't know the amount that will be allocated)

pkg/types/node_resources_status.go
gshaibi: You're right. thanks.

cmd/util/gpu.go
omer-dayan: Then maybe lets merge those:
	WorkloadCurrentAllocatedGPUs       = "runai-current-allocated-gpus"
	WorkloadCurrentAllocatedGPUsMemory = "runai-current-allocated-gpus-memory"
	PodAllocatedMigGPUs                = "runai-allocated-mig-gpus"

Arent they say the same thing?

gshaibi: But the first 2 are on the Job's level, and the latter is on the Pod one.
You want me to refactor the first 2 to be on the Pod level?

omer-dayan: No I guess we will need to do it in a ticket of itself

omer-dayan (approved):

PR #157, opened 2021-07-26
Added base directory for release tars

Gil-AdB:


enoodle:
@omer-dayan shouldn't we adjust the install script as well?

PR #171, opened 2021-08-09
RUN-1710/template support cli

oferla:


cmd/template/describe.go
```diff
@@ -192,21 +206,6 @@ func describeTemplatePrint(w io.Writer, template api.JobTemplate) {
 	}
 }
 
-//
-//   a utility function to return the set of keys of a map[string]whatever, sorted
-//
```
oferla, line 197: Move this function to utils instead of hiding it here.

itayvallach (approved):

PR #193, opened 2021-10-07
Run 1770 add mig profile autocomplete

natasharomm:


enoodle (approved):

enoodle (approved):

cmd/util/node.go
```diff
@@ -73,4 +78,24 @@ func IsNodeReady(node v1.Node) bool {
...
+	}
+
+	nodeList, err := kubeClient.GetClientset().CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	nodes := []v1.Node{}
+	for _, node := range nodeList.Items {
```
omer-dayan, line 95: Why not return nodeList.Items?

cmd/util/mig.go
```diff
@@ -11,3 +32,38 @@ func IsMigInstanceAnnotation(key string) bool {
 func ConvertMigResourceFullNameToMigInstance(migResourceFullName string) string {
 	return strings.TrimPrefix(migResourceFullName, migResourcePrefix)
 }
+
+func GetExistingMigProfiles() map[string]bool {
+	nodes := GetAllNodes()
+	availableMigProfilesMap := map[string]bool{}
+
+	for _, node := range nodes {
+		if !IsNodeReady(node) {
```
omer-dayan, line 41: Why like that? If node is not in ready state we dont want him to be able to autocomplete the mig profile?

cmd/job/submit/requested_gpu_handler.go
```diff
@@ -2,41 +2,36 @@ package submit
...
 	}
 
-	if submitArgs.MigDevice != "" {
-		if _, ok := supportedMigDevices[submitArgs.MigDevice]; !ok {
-			return fmt.Errorf("unsupported mig device: %v", submitArgs.MigDevice)
+	if submitArgs.MigProfile != "" {
+		existingMigProfiles := util.GetExistingMigProfiles()
+
+		if _, ok := existingMigProfiles[submitArgs.MigProfile]; !ok {
+			return fmt.Errorf("mig profile: '%v' doesn't exist in the system", submitArgs.MigProfile)
```
omer-dayan, line 32: In my opinion we should not do it like that. I will talk to Omer about it. We still should be able to submit the job if the mig-profile does not exist in the cluster.

However we need to validate the mig-profile is valid one

cmd/util/mig.go
natasharomm: yes, that's what OmerB described in the ticket.

cmd/util/node.go
natasharomm: it's possible, but the caller doesn't really need all the extra information in "NodeList", the caller only needs a list of the nodes - []Node.

PR #200, opened 2021-11-07
RUN-2380 Support kubectl versions with special characters

gshaibi:


pkg/util/kubectl/kubectl.go
```diff
@@ -60,12 +61,13 @@ func SupportOldDryRun() (bool, error) {
 		return false, fmt.Errorf("%s%w", errPrefix, err)
 	}
 
-	majorVersion, err := strconv.Atoi(response.ClientVersion.Major)
+	reg := regexp.MustCompile("[0-9]+")
```
omer-dayan, line 64: This Regex - "[0-9]+" has returned to me the number even when there was a lot of string, like: "25+vmware". Thats by design, right?

omer-dayan (approved):

pkg/util/kubectl/kubectl.go
gshaibi: yes

PR #203, opened 2021-11-15
RUN-2483 MPI CRD validation

gshaibi:


cmd/job/submit/submit_mpi.go
```diff
@@ -162,8 +165,28 @@ func (submitArgs submitMPIJobArgs) check() error {
 	return nil
 }
 
+func isMpiJobCrdExist() (bool, error) {
+	crds, err := kubectl.GetCRDs()
```
omer-dayan, line 169: Why run 'kubectl get crds' and then check in the list and not run 'kubectl get crd mpijobs'?

pkg/util/kubectl/kubectl.go
```diff
@@ -413,6 +413,11 @@ func PortForward(ports []string, PortForwardAddr string, serviceName string, nam
 	return kubectlAttched(args)
 }
 
+func GetCRDs() (string, error) {
+	args := []string{"get", "customresourcedefinitions.apiextensions.k8s.io"}
+	return kubectl(args)
```
omer-dayan, line 418: We much prefer use client-go instead of direct kubectl commands

gshaibi: Couldn't find API for CRDs

cmd/job/submit/submit_mpi.go
gshaibi: Wanted to generalize it.
I'll change it.

pkg/util/kubectl/kubectl.go
omer-dayan: You can use unstructured dynamic client

pkg/util/kubectl/kubectl.go
```diff
@@ -413,6 +413,21 @@ func PortForward(ports []string, PortForwardAddr string, serviceName string, nam
 	return kubectlAttched(args)
 }
 
+func IsCrdExist(crdName string) (bool, error) {
```
omer-dayan, line 416: No need

omer-dayan (approved):

PR #205, opened 2021-11-16
changed the way supplimentary groups are decoded for user

romanbaron:
Groups attribute can be defined in KeyCloak as multivalued int attribute (with ## delimiter):

Screen Shot 2021-11-16 at 17 20 16

In this case it is automatically decoded into a slice of ints:

Screen Shot 2021-11-16 at 17 17 32

So, groupsSliceAtoi is no longer required.


omer-dayan (commented):
Thats great news.
However isnt it breaking the API?

What will happened to all those who uses this feature and their KC is sending strings? 
Will that stop working once the upgrade the CLI?

romanbaron:
@omer-dayan - I am done changing the code according to your instructions, now SupplementaryGroups and Groups are read separately from the token. Groups will only be read if SupplementaryGroups are empty.

omer-dayan (approved):

PR #206, opened 2021-11-22
when sso is enabled, will use email as client subject claim field

romanbaron:


romanbaron:
Will be documented and done manually. Closing.

PR #209, opened 2021-12-15
Removed dependency in kubectl and helm

omeryahud:
`runai` cli tool will now template and submit jobs directly against
kube-api instead of templating with `helm` and submitting with `kubectl`


omeryahud:
This is ready for review, will be tested using the existing QA E2E test suite

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+
+	for i, pvcVol := range pvcStrs {
+		pvcs = append(pvcs, *structPvc(pvcVol, jobName, i))
+	}
+
+	return pvcs
+}
+
+func structPvc(pvcStr, jobName string, index int) *v1.PersistentVolumeClaim {
+	split := strings.Split(pvcStr, ":")
```
oferla, line 676: better call it "parts"... split is the action not the result.

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+	if accessMode == AccessModeReadOnly {
+		volumeMount.ReadOnly = true
+	}
+
+	return volumeMount
+}
+
+func parsePvcVolumeMount(pvcVolStr, namePrefix string, index int) *v1.VolumeMount {
+	volumeMount := &v1.VolumeMount{}
+	splits := strings.Split(pvcVolStr, ":")
```
oferla, line 765: We moved from "split" to "splits"?!! not even sure it's correct English... parts. much better name.

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+		if accessMode == AccessModeReadOnly {
+			volumeMount.ReadOnly = true
+		}
+	} else if len(splits) == 3 {
+		accessMode := splits[2]
+
+		if accessMode == AccessModeReadOnly {
+			volumeMount.ReadOnly = true
+		}
+	}
```
oferla, line 789: if you sure len(splits) is either 3 or 4 no need for the "else if" just "else".
if you're not sure add "else" with error message

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+			return nil, err
+		}
+
+		containerPort.ContainerPort = int32(intPort)
+	}
+
+	return containerPort, nil
+}
+
+func portForward(localAddresses, ports []string, jobName, namespace string,
```
oferla, line 900: Is the code of this function copied from somewhere (like kubectl?) I could have hoped there is a standard library function for this... otherwise not sure this is the place for it. better move to util of some sort. seems too low level for here.

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+	kubeClient *client.Client) (stop chan<- struct{}, errC <-chan error, err error) {
+	restConfig, _, err := client.GetRestConfig()
+	if err != nil {
+		return nil, nil, err
+	}
```
oferla, line 905: Just to verify: are you sure you're using the same config as the CLI itself?

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+				stopChans[port] = make(chan struct{})
+				readyChans[port] = make(chan struct{})
+			}
+		}(port, portForwardReq)
+	}
+
+	return globalStopChan, errChan, nil
+}
+
+func coreV1TypeMeta(obj runtime.Object) metav1.TypeMeta {
```
oferla, line 981: Again not sure that all the typing functions should be here. would move them to a more isolated places than this file, which should include the submit business logic.

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+			return err
+		}
+
+		dynamicClient := kubeClient.GetDynamicClient()
+
+		gvk := res.GetObjectKind().GroupVersionKind()
+		gvs := schema.GroupVersionResource{
+			Group:    strings.ToLower(gvk.Group),
+			Version:  strings.ToLower(gvk.Version),
+			Resource: pluralize.NewClient().Pluralize(strings.ToLower(gvk.Kind), 2, false),
```
oferla, line 1040: 2? maybe constant whose name can explain what is this miraculous 2?

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func getPodLogs(pod *v1.Pod, kubeClient *client.Client) (string, error) {
```
oferla, line 1057: not the right place for this function (I think it is not even being used in this source)

cmd/job/submit/submit.go
```diff
@@ -531,3 +602,479 @@ func AlignArgsPreParsing(args []string) []string {
...
+	logsRes := logsReq.Do(context.TODO())
+	logsBuf, err := logsRes.Raw()
+	if err != nil {
+		return "", err
+	}
+
+	return string(logsBuf), nil
+}
+
+func mergeStringMaps(a, b map[string]string) map[string]string {
```
oferla, line 1068: again - looking at this source file, I an not seeing this function used. maybe the browser find operation is malfunctioning... could be. but if not used better be removed and if used better be moved.
ושלמה ארצי אפילו יכול להלחין את המשפט הזה

cmd/job/submit/submit_mpi.go
```diff
@@ -47,6 +55,12 @@ runai submit-mpi --name distributed-job --processes=2 -g 1 \
 	-i gcr.io/run-ai-demo/quickstart-distributed`
 )
 
+const (
```
oferla, line 58: worth adding comment explaining the two numeric constants.

cmd/job/submit/submit_runai.go
```diff
@@ -123,14 +119,15 @@ func NewRunaiJobCommand() *cobra.Command {
 				submitArgs.Completions = &interactiveCompletions
 			}
 
-			err = submitRunaiJob(submitArgs, clientset, *runaijobClient)
-			if err != nil {
-				fmt.Println(err)
+			if len(submitArgs.Image) == 0 {
```
oferla, line 122: isn't it too early to test this? I hope that in here the submitArgs already have all the defaults from jobsettings and user/global templates merged to it.

cmd/job/submit/submit_runai.go
```diff
@@ -521,3 +542,561 @@ func substituteEnvVariables(envVars []string) {
...
+		}
+
+		svc.Spec.Ports = append(svc.Spec.Ports, *svcPort)
+	}
+
+	return svc, nil
+}
+
+func parseServicePortLiterals(portStr string) (strPort, strTargetPort string) {
+	slice := strings.Split(portStr, ":")
```
oferla, line 1064: I still think parts is better 🥇

cmd/job/submit/submit_runai.go
```diff
@@ -521,3 +542,561 @@ func substituteEnvVariables(envVars []string) {
...
+		},
+	}
+
+	if len(submitArgs.Ports) == 1 {
+		ingressPort := parseSingleIngressRulePath(submitArgs.Ports[0], submitArgs.Project, submitArgs.Name)
+		ingress.Spec.Rules[0].HTTP.Paths = append(ingress.Spec.Rules[0].HTTP.Paths, *ingressPort)
+	} else {
+		for _, port := range submitArgs.Ports {
+			ingressPort := parseIngressRulePath(port, submitArgs.Project, submitArgs.Name)
+			ingress.Spec.Rules[0].HTTP.Paths = append(ingress.Spec.Rules[0].HTTP.Paths, *ingressPort)
```
oferla, line 973: Everything is squeezed to rule 0? no typo here?

cmd/job/submit/submit_runai.go
```diff
@@ -521,3 +542,561 @@ func substituteEnvVariables(envVars []string) {
...
+
+		}
+	}
+
+	return ingress
+}
+
+func parseIngressRulePath(portStr, project, jobName string) *networkingv1.HTTPIngressPath {
+	split := strings.Split(portStr, ":")
```
oferla, line 982: finally you had the slices which is better than split, and now you're back to split?!!?! 👎

cmd/job/submit/submit_runai.go
```diff
@@ -521,3 +542,561 @@ func substituteEnvVariables(envVars []string) {
...
+	port := split[0]
+
+	portIntStr := intstr.FromString(port)
+	portValue := int32((&portIntStr).IntValue())
```
oferla, line 986: I hope that go has a better way to cover port from string to int32...

cmd/job/submit/submit_runai.go
```diff
@@ -521,3 +542,561 @@ func substituteEnvVariables(envVars []string) {
...
+}
+
+func parseSingleIngressRulePath(portStr, project, jobName string) *networkingv1.HTTPIngressPath {
+	// This is very similar to parseIngressRulePath, I decided to implement it in order
+	// to align with the helm template it replaces, and not change the behavior
+	split := strings.Split(portStr, ":")
+	port := split[0]
+
+	portIntStr := intstr.FromString(port)
+	portValue := int32((&portIntStr).IntValue())
```
oferla, line 1009: same comment again.

cmd/job/submit/submit_runai.go
```diff
@@ -521,3 +542,561 @@ func substituteEnvVariables(envVars []string) {
...
+			strPort = slice[1]
+			strTargetPort = slice[1]
+		} else if slice[1] == "" {
+			strPort = slice[0]
+			strTargetPort = slice[0]
+		} else {
+			strPort = slice[0]
+			strTargetPort = slice[1]
+		}
+	}
```
oferla, line 1080: Maybe a one lines comment explaining the logic here?

pkg/portforward/portforward.go
```diff
@@ -0,0 +1,61 @@
...
+	// ReadyCh communicates when the tunnel is ready to receive traffic
+	ReadyCh chan struct{}
+	// LocalAddresses is a list of local addresses to listen on
+	LocalAddresses []string
+}
+
+func PortForwardAPod(req PortForwardAPodRequest) error {
+	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward",
+		req.Pod.Namespace, req.Pod.Name)
+	hostIP := strings.TrimLeft(req.RestConfig.Host, "htps:/")
```
oferla, line 38: Only because it's 11pm I really did google on "htps" but since it did not find anything related I assume this is a typo and should be https:, but maybe you'll surprise me tomorrow...

pkg/portforward/portforward.go
```diff
@@ -0,0 +1,61 @@
...
+
+	transport, upgrader, err := spdy.RoundTripperFor(req.RestConfig)
+	if err != nil {
+		return err
+	}
+
+	dialer := spdy.NewDialer(upgrader,
+		&http.Client{Transport: transport},
+		http.MethodPost,
+		&url.URL{Scheme: "https", Path: path, Host: hostIP})
```
oferla, line 48: ...or maybe not.

pkg/workflow/submit.go
```diff
@@ -126,7 +89,7 @@ func submitConfigMap(name, namespace string, generateSuffix bool, clientset kube
 
 func createEmptyConfigMap(name, baseName, namespace string, index int, clientset kubernetes.Interface) (*corev1.ConfigMap, error) {
 	labels := make(map[string]string)
-	labels[kubectl.JOB_CONFIG_LABEL_KEY] = kubectl.JOB_CONFIG_LABEL_VALUES
+	//labels[kubectl.JOB_CONFIG_LABEL_KEY] = kubectl.JOB_CONFIG_LABEL_VALUES
```
oferla, line 92: Dead code?

oferla (changes requested):

cmd/job/submit/submit.go
omeryahud: len(*slice*) can also be 2:
`--pvc stringArray      Mount a persistent volume. Syntax: 'StorageClass[optional]:Size:ContainerMountPath:ro[optional]'`

omeryahud: But something else looks strange to me, I'll look into it

omeryahud: Never mind, it cannot be 2, only 3 or 4

cmd/job/submit/submit.go
omeryahud: It is not copied from `kubectl`, the function `portforward.PortForwardAPod` uses the k8s libraries to perform port forwarding

cmd/job/submit/submit.go
omeryahud: Yes, any caller to `client.GetRestConfig` will be configured to access to same cluster context

cmd/job/submit/submit.go
omeryahud: Done, created `*_resources.go` files for each job

cmd/job/submit/submit.go
omeryahud: It is being used, but I agree that I can move it to a different place

cmd/job/submit/submit_mpi.go
omeryahud: I wish I knew what `SlotsPerWorker` means

cmd/job/submit/submit_runai.go
omeryahud: Yes, you can see here that templates are handled prior to this check https://github.com/run-ai/runai-cli/pull/209/files/8440bf91cf5b6a43e70a0e761bc4f226dd800e4b#diff-99a1c3ae9f63e0a4684aa8935157ab4b8ad767323d9d0a07194f9ee4c97573bcR91

cmd/job/submit/submit_runai.go
omeryahud: No typo, everything is in the same rule https://github.com/run-ai/runai-cli/pull/209/files/8440bf91cf5b6a43e70a0e761bc4f226dd800e4b?authenticity_token=IKGx%2F8eYDFbF54%2FvwKQRfniopfNCXScSzH9uOfqM8F%2FEuvrCHRrlrLS5dpSeoO3Z3XDskH54gEiEBKG5Jhq20A%3D%3D&file-filters%5B%5D=.go&file-filters%5B%5D=.mod&file-filters%5B%5D=.sum&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.tpl&file-filters%5B%5D=.yaml&file-filters%5B%5D=dotfile#diff-99e59f58a53e00378ebf58f41338f7c7f9bba222673e8fa6f0a42b852f6bc32aL19

cmd/job/submit/submit_runai.go
omeryahud: This is not a go thing, it's a k8s thing, the API defines a struct called `IntOrString`, so no better way for that

cmd/job/submit/submit_runai.go
omeryahud: Same as above

cmd/job/submit/submit_runai.go
omeryahud: Done, added a comment at the top of the function

cmd/job/submit/submit.go
oferla: I think it should not reside in submit.go, but this is outside the scope for this task we can do it somewhere else. Should be a file handling port forwarding, waiting for pod to run and then log, etc... it is all outside the scope for submit.go, but not mandatory for this PR if it is too much work.

itayvallach (approved):

oferla (approved):

itayvallach (approved):

PR #211, opened 2021-12-22
RUN-2663-Fixed-ALLOCATED-GPUs-for-fractions

natasharomm:


gshaibi (approved):

pkg/nodes/node_info.go
```diff
@@ -144,6 +153,11 @@ func calcNodePodsResourcesStatus(ni *NodeInfo) *types.PodResourcesStatus {
 	return &totalPodsResourcesStatus
 }
 
+func numOfSharedGpus(ni *NodeInfo) int {
```
gshaibi, line 156: I think we should define the term "shared GPU". 
Here you mean fractional/swap, but I think in other places, MIG is considered shared as well.

gshaibi (approved):

PR #220, opened 2022-02-09
add support for mig A100 80GB

enoodle:


cmd/job/submit/requested_gpu_handler.go
```diff
@@ -26,10 +27,6 @@ func handleRequestedGPUs(submitArgs *submitArgs) error {
 	}
 
 	if submitArgs.MigProfile != "" {
-		if _, ok := util.SupportedMigProfiles[util.A100]; !ok {
```
galbenyair, line 29: don't we need this validation?

enoodle: This validation was not dependent on the user input at all, so this was always the same..

enoodle: `util.SupportedMigProfiles` is constant, `util.A100` is constant.. this is always the same.

omer-dayan (approved):

PR #222, opened 2022-02-21
fix missing gcloud dependancy in 'push' job

romanbaron:


romanbaron:
will open with a proper branch name and jira ticket

PR #225, opened 2022-02-28
RUN-1606 - Consistent sorting for runai top job

omric-runai:
Implemented consisted sorting for JobView slice used in `runai top job`. Sorting is done lexicographically by project and then by job name (after consulting with Yaron), but the implementation allows us to change the logic pretty easily.


pkg/types/ordered_job_view.go
```diff
@@ -0,0 +1,20 @@
...
+func (job OrderedJobView) Len() int {
+	return len(job)
+}
+
+func (job OrderedJobView) Less(i, j int) bool {
+	if job[i].Info.Project == job[j].Info.Project {
+		return job[i].Info.Name < job[j].Info.Name
+	}
+
+	return job[i].Info.Project < job[j].Info.Project
```
romanbaron, line 15: you can assign job[i/j].Info to some variable, it is a very small change but can make it a bit more readable

natasharomm (approved):

PR #226, opened 2022-02-28
Bugfix - Change output of login from stderr to stdout

naomi-friedrich:
In the past, when you ran `runai login` it would have print output of "Username: " and "Password: " and send it to `Stderr`, instead of `Stdout` which is the correct way in linux standard.

We have found this issue through automation process


itayvallach:
😂😂😂😂

pkg/authentication/flows/password/password.go
```diff
@@ -133,7 +133,7 @@ func getRawCredentials() (string, string, error) {
 }
 
 func readString(prompt string) (string, error) {
-	if _, err := fmt.Fprint(os.Stderr, prompt); err != nil {
+	if _, err := fmt.Fprint(os.Stdout, prompt); err != nil {
```
omer-dayan, line 136: Dont you think we need to refactor all the logic?

omer-dayan: Did you think about utilization?

omric-runai: I think we should consider adding about 3-4 utilizations in here

enoodle:
Thank you for your contribution @naomi-friedrich

pkg/authentication/flows/password/password.go
oferla: need to check if os.Stdout is not nil.

oferla (approved):

natasharomm (approved):

